PR #112, opened 2022-11-15
Rollover vault

aalavandhan:
- Users deposit AMPL for vault notes
- Vault "deploys" ampl by tranching and performing rollover, it "recovers" AMPL by redeeming mature tranches
- On redemption users get a slice of all the assets held by the vault

At any given time the vault could hold, the underlying asset (deposit asset AMPL), the earned asset (SPOT) and the deployed assets (tranches).

### DIFF
--- a/spot-contracts/contracts/_utils/BondHelpers.sol
+++ b/spot-contracts/contracts/_utils/BondHelpers.sol
@@ -191,27 +191,33 @@ library BondHelpers {
         return (td, collateralBalances, trancheSupplies);
     }
 
-    /// @notice Given a bond, retrieves the collateral redeemable for
-    ///         each tranche held by the given address.
+    /// @notice Given a bond and a user address computes the tranche amounts proportional to the tranche ratios,
+    //          that can be redeemed for the collateral before maturity.
     /// @param b The address of the bond contract.
     /// @param u The address to check balance for.
-    /// @return The tranche data and an array of collateral balances.
-    function getTrancheCollateralBalances(IBondController b, address u)
+    /// @return The tranche data and an array of tranche token balances.
+    function computeRedeemableTrancheAmounts(IBondController b, address u)
         internal
         view
         returns (TrancheData memory, uint256[] memory)
     {
-        TrancheData memory td;
-        uint256[] memory collateralBalances;
-        uint256[] memory trancheSupplies;
+        TrancheData memory td = getTrancheData(b);
+        uint256[] memory redeemableAmts = new uint256[](td.trancheCount);
 
-        (td, collateralBalances, trancheSupplies) = getTrancheCollateralizations(b);
-
-        uint256[] memory balances = new uint256[](td.trancheCount);
-        for (uint8 i = 0; i < td.trancheCount; i++) {
-            balances[i] = (td.tranches[i].balanceOf(u) * collateralBalances[i]) / trancheSupplies[i];
+        // We calculate the minimum value of {trancheBal/trancheRatio} across tranches
+        uint256 min = type(uint256).max;
+        uint8 i;
+        for (i = 0; i < td.trancheCount && min != 0; i++) {
+            uint256 d = (td.tranches[i].balanceOf(u) * TRANCHE_RATIO_GRANULARITY) / td.trancheRatios[i];
+            if (d < min) {
+                min = d;
+            }
         }
 
-        return (td, balances);
+        for (i = 0; i < td.trancheCount; i++) {
+            redeemableAmts[i] = (td.trancheRatios[i] * min) / TRANCHE_RATIO_GRANULARITY;
+        }
+
+        return (td, redeemableAmts);
     }
 }
--- a/spot-contracts/contracts/_utils/BondHelpers.sol
+++ b/spot-contracts/contracts/_utils/BondHelpers.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.18;
 
 import { SafeCastUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/math/SafeCastUpgradeable.sol";
+import { MathUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/math/MathUpgradeable.sol";
 
 import { IERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
 import { IBondController } from "../_interfaces/buttonwood/IBondController.sol";
@@ -66,6 +67,9 @@ library TrancheHelpers {
  *
  */
 library BondHelpers {
+    using SafeCastUpgradeable for uint256;
+    using MathUpgradeable for uint256;
+
     // Replicating value used here:
     // https://github.com/buttonwood-protocol/tranche/blob/main/contracts/BondController.sol
     uint256 private constant TRANCHE_RATIO_GRANULARITY = 1000;
@@ -92,7 +96,7 @@ library BondHelpers {
     /// @return The tranche data.
     function getTrancheData(IBondController b) internal view returns (TrancheData memory) {
         TrancheData memory td;
-        td.trancheCount = SafeCastUpgradeable.toUint8(b.trancheCount());
+        td.trancheCount = b.trancheCount().toUint8();
         td.tranches = new ITranche[](td.trancheCount);
         td.trancheRatios = new uint256[](td.trancheCount);
         // Max tranches per bond < 2**8 - 1
@@ -191,8 +195,8 @@ library BondHelpers {
         return (td, collateralBalances, trancheSupplies);
     }
 
-    /// @notice Given a bond and a user address computes the tranche amounts proportional to the tranche ratios,
-    //          that can be redeemed for the collateral before maturity.
+    /// @notice For a given bond and user address, computes the maximum number of each of the bond's tranches
+    ///         the user is able to redeem before the bond's maturity. These tranche amounts necessarily match the bond's tranche ratios.
     /// @param b The address of the bond contract.
     /// @param u The address to check balance for.
     /// @return The tranche data and an array of tranche token balances.
@@ -204,18 +208,24 @@ library BondHelpers {
         TrancheData memory td = getTrancheData(b);
         uint256[] memory redeemableAmts = new uint256[](td.trancheCount);
 
-        // We calculate the minimum value of {trancheBal/trancheRatio} across tranches
-        uint256 min = type(uint256).max;
+        // Calculate how many underlying assets could be redeemed from each tranche balance,
+        // assuming other tranches are not an issue, and record the smallest amount.
+        uint256 minUnderlyingOut = type(uint256).max;
         uint8 i;
-        for (i = 0; i < td.trancheCount && min != 0; i++) {
-            uint256 d = (td.tranches[i].balanceOf(u) * TRANCHE_RATIO_GRANULARITY) / td.trancheRatios[i];
-            if (d < min) {
-                min = d;
+        for (i = 0; i < td.trancheCount; i++) {
+            uint256 d = td.tranches[i].balanceOf(u).mulDiv(TRANCHE_RATIO_GRANULARITY, td.trancheRatios[i]);
+            if (d < minUnderlyingOut) {
+                minUnderlyingOut = d;
+            }
+
+            // if one of the balances is zero, we return
+            if (minUnderlyingOut == 0) {
+                return (td, redeemableAmts);
             }
         }
 
         for (i = 0; i < td.trancheCount; i++) {
-            redeemableAmts[i] = (td.trancheRatios[i] * min) / TRANCHE_RATIO_GRANULARITY;
+            redeemableAmts[i] = td.trancheRatios[i].mulDiv(minUnderlyingOut, TRANCHE_RATIO_GRANULARITY);
         }
 
         return (td, redeemableAmts);
--- a/spot-contracts/contracts/vaults/RolloverVault.sol
+++ b/spot-contracts/contracts/vaults/RolloverVault.sol
@@ -217,7 +217,7 @@ contract RolloverVault is
         uint256 amount;
     }
 
-    /// @notice Burns notes and sends a proportional share vault's assets back to {msg.sender}.
+    /// @notice Burns notes and sends a proportional share of vault's assets back to {msg.sender}.
     /// @param notes The amount of notes to be burnt.
     /// @return The list of asset tokens and amounts redeemed.
     function redeem(uint256 notes) external nonReentrant whenNotPaused returns (TokenAmount[] memory) {
@@ -293,7 +293,7 @@ contract RolloverVault is
     /// @notice Checks if the given token is held by the vault.
     /// @param token The address of a token to check.
     function isVaultAsset(IERC20Upgradeable token) public view returns (bool) {
-        return token == underlying || _deployed.contains(address(token)) || address(perp) == address(token);
+        return (token == underlying) || _deployed.contains(address(token)) || (address(perp) == address(token));
     }
 
     /// @notice Total count of deployed asset tokens held by the vault.
--- a/spot-contracts/contracts/vaults/RolloverVault.sol
+++ b/spot-contracts/contracts/vaults/RolloverVault.sol
@@ -14,6 +14,7 @@ import { TrancheData, TrancheHelpers, BondHelpers } from "../_utils/BondHelpers.
 import { IERC20Upgradeable, IPerpetualTranche, IBondIssuer, IBondController, ITranche } from "../_interfaces/IPerpetualTranche.sol";
 
 // TODO: create a IVault interface
+// TODO: add mint cap
 
 /// @notice Expected asset to be a vault asset.
 /// @param token Address of the token.
--- a/spot-contracts/contracts/vaults/RolloverVault.sol
+++ b/spot-contracts/contracts/vaults/RolloverVault.sol
@@ -427,8 +427,8 @@ contract RolloverVault is
             ITranche tranche = ITranche(_deployed.at(i));
             IBondController bond = IBondController(tranche.bond());
 
-            // if bond has mature, redeem the tranche token
-            if (bond.timeToMaturity() == 0) {
+            // if bond has matured, redeem the tranche token
+            if (bond.timeToMaturity() <= 0) {
                 if (!bond.isMature()) {
                     bond.mature();
                 }
--- a/spot-contracts/contracts/vaults/RolloverVault.sol
+++ b/spot-contracts/contracts/vaults/RolloverVault.sol
@@ -16,7 +16,7 @@ import { IERC20Upgradeable, IPerpetualTranche, IBondIssuer, IBondController, ITr
 // TODO: create a IVault interface
 // TODO: add mint cap
 
-/// @notice Expected asset to be a vault asset.
+/// @notice Expected asset to be a valid vault asset.
 /// @param token Address of the token.
 error UnexpectedAsset(IERC20Upgradeable token);
 
@@ -134,6 +134,7 @@ contract RolloverVault is
         underlying = perp_.collateral();
         _syncAsset(underlying);
 
+        assert(underlying != perp_);
         perp = perp_;
     }
 
@@ -350,61 +351,65 @@ contract RolloverVault is
         //       there after the list is NOT ordered by maturity.
         IERC20Upgradeable[] memory rolloverTokens = perp_.getReserveTokensUpForRollover();
 
-        // Batch rollover:
-        // For each tranche token (held by the vault) to be rolled over,
-        // and each tranche token up for rollover (held by perp),
-        // Execute rollover and move on to the next pair.
+        // Batch rollover
         uint256 totalPerpRolledOver = 0;
-        uint8 i = 0;
-        uint256 j = 0;
-        while (i < td.trancheCount && j < rolloverTokens.length) {
-            ITranche trancheIn = td.tranches[i];
-            IERC20Upgradeable tokenOut = rolloverTokens[j];
+        uint8 vaultTokenIdx = 0;
+        uint256 perpTokenIdx = 0;
 
-            // Compute available tranche in
-            uint256 trancheInAmtAvailable = trancheIn.balanceOf(address(this));
+        // We pair tranche tokens held by the vault with tranche tokens held by perp,
+        // And execute the rollover and continue to the next token with a usable balance.
+        while (vaultTokenIdx < td.trancheCount && perpTokenIdx < rolloverTokens.length) {
+            // trancheIntoPerp refers to the tranche going into perp from the vault
+            ITranche trancheIntoPerp = td.tranches[vaultTokenIdx];
 
-            // trancheIn tokens are exhausted
-            if (trancheInAmtAvailable == 0) {
-                // Rollover is a no-op, so skipping to next trancheIn
-                i++;
+            // tokenOutOfPerp is the reserve token coming out of perp into the vault
+            IERC20Upgradeable tokenOutOfPerp = rolloverTokens[perpTokenIdx];
+
+            // compute available token out
+            uint256 tokenOutAmtAvailable = address(tokenOutOfPerp) != address(0)
+                ? tokenOutOfPerp.balanceOf(perp_.reserve())
+                : 0;
+
+            // trancheIntoPerp tokens are NOT exhausted but tokenOutOfPerp is exhausted
+            if (tokenOutAmtAvailable == 0) {
+                // Rollover is a no-op, so skipping to next tokenOutOfPerp
+                perpTokenIdx++;
                 continue;
             }
 
-            // compute available token out
-            uint256 tokenOutAmtAvailable = address(tokenOut) != address(0) ? tokenOut.balanceOf(perp_.reserve()) : 0;
+            // Compute available tranche in
+            uint256 trancheInAmtAvailable = trancheIntoPerp.balanceOf(address(this));
 
-            // trancheIn tokens are NOT exhausted but tokenOut is exhausted
-            if (tokenOutAmtAvailable == 0) {
-                // Rollover is a no-op, so skipping to next tokenOut
-                j++;
+            // trancheInAmtAvailable is exhausted
+            if (trancheInAmtAvailable == 0) {
+                // Rollover is a no-op, so skipping to next trancheIntoPerp
+                vaultTokenIdx++;
                 continue;
             }
 
             // Preview rollover
             IPerpetualTranche.RolloverPreview memory rd = perp_.computeRolloverAmt(
-                trancheIn,
-                tokenOut,
+                trancheIntoPerp,
+                tokenOutOfPerp,
                 trancheInAmtAvailable,
                 tokenOutAmtAvailable
             );
 
-            // trancheIn isn't accepted by perp, likely because yield=0
+            // trancheIntoPerp isn't accepted by perp, likely because it's yield=0, refer perp docs for more info
             if (rd.perpRolloverAmt == 0) {
-                // Though `trancheInAmtAvailable` it is unusable so marking as exhausted
-                // Rollover is a no-op, so skipping to next trancheIn
-                i++;
+                // Rollover is a no-op, so skipping to next trancheIntoPerp
+                vaultTokenIdx++;
                 continue;
             }
 
             // Perform rollover
-            _checkAndApproveMax(trancheIn, address(perp_), trancheInAmtAvailable);
-            perp_.rollover(trancheIn, tokenOut, trancheInAmtAvailable);
+            _checkAndApproveMax(trancheIntoPerp, address(perp_), trancheInAmtAvailable);
+            perp_.rollover(trancheIntoPerp, tokenOutOfPerp, trancheInAmtAvailable);
 
             // sync holdings
-            _syncDeployedAsset(trancheIn);
-            if (tokenOut != underlying) {
-                _syncDeployedAsset(tokenOut);
+            _syncDeployedAsset(trancheIntoPerp);
+            if (tokenOutOfPerp != underlying) {
+                _syncDeployedAsset(tokenOutOfPerp);
             }
             _syncAsset(perp_);
             _syncAsset(underlying);
--- a/spot-contracts/contracts/vaults/RolloverVault.sol
+++ b/spot-contracts/contracts/vaults/RolloverVault.sol
@@ -152,6 +152,7 @@ contract RolloverVault is
         _unpause();
     }
 
+    /// @notice Transfers a non-vault token out of the contract, which may have been added accidentally.
     /// @param token The token address.
     /// @param to The destination address.
     /// @param amount The amount of tokens to be transferred.
@@ -225,9 +226,6 @@ contract RolloverVault is
         uint256 deployedCount_ = _deployed.length();
         uint256 assetCount = 2 + deployedCount_;
 
-        // burn notes
-        _burn(_msgSender(), notes);
-
         // aggregating vault assets to be redeemed
         TokenAmount[] memory redemptions = new TokenAmount[](assetCount);
         redemptions[0].token = underlying;
@@ -235,6 +233,9 @@ contract RolloverVault is
             redemptions[i + 1].token = IERC20Upgradeable(_deployed.at(i));
         }
         redemptions[deployedCount_ + 1].token = IERC20Upgradeable(perp);
+
+        // burn notes
+        _burn(_msgSender(), notes);
 
         // calculating amounts and transferring assets out proportionally
         for (uint256 i = 0; i < assetCount; i++) {
--- a/spot-contracts/contracts/vaults/RolloverVault.sol
+++ b/spot-contracts/contracts/vaults/RolloverVault.sol
@@ -222,25 +222,21 @@ contract RolloverVault is
     /// @return The list of asset tokens and amounts redeemed.
     function redeem(uint256 notes) external nonReentrant whenNotPaused returns (TokenAmount[] memory) {
         uint256 totalNotes = totalSupply();
-        uint256 deployedCount_ = deployedCount();
-        uint256 earnedCount_ = earnedCount();
-        uint256 assetCount = 1 + deployedCount_ + earnedCount_;
-
-        TokenAmount[] memory redemptions = new TokenAmount[](assetCount);
+        uint256 deployedCount_ = _deployed.length();
+        uint256 assetCount = 2 + deployedCount_;
 
         // burn notes
         _burn(_msgSender(), notes);
 
         // aggregating vault assets to be redeemed
+        TokenAmount[] memory redemptions = new TokenAmount[](assetCount);
         redemptions[0].token = underlying;
         for (uint256 i = 0; i < deployedCount_; i++) {
-            redemptions[i + 1].token = deployedAt(i);
+            redemptions[i + 1].token = IERC20Upgradeable(_deployed.at(i));
         }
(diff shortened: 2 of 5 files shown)


spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+    /// @notice Recovers deployed funds.
+    function recover() public nonReentrant whenNotPaused {
+        _redeemTranches();
+    }
+
+    /// @notice Deposits the provided asset from {msg.sender} into the vault and mints notes.
+    /// @param token The address of the asset to deposit.
+    /// @param amount The amount tokens to be deposited into the vault.
+    /// @return The amount of notes.
+    function deposit(IERC20Upgradeable token, uint256 amount) external nonReentrant whenNotPaused returns (uint256) {
```
nms-7, line 197: consider removing the `token` param entirely

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,525 @@
...
+        TrancheData memory td = bond.getTrancheData();
+        for (uint8 i = 0; i < td.trancheCount; i++) {
+            _syncDeployedAsset(td.tranches[i]);
+        }
+        _syncAsset(underlying);
+    }
+
+    /// @dev Rolls over freshly tranched tokens from the given bond for older tranches (close to maturity) from perp.
+    /// @return The amount of perps rolled over.
+    function _rollover(IPerpetualTranche perp_, IBondController depositBond) private returns (uint256) {
```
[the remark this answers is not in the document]
aalavandhan: Eventually if we have a Z-perp, this vault could perform rollovers on both. Didn't want to add that logic yet but we can always upgrade down the road.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+            // trancheIn isn't accepted by perp, likely because yield=0
+            if (rd.perpRolloverAmt == 0) {
+                // Though `trancheInAmtAvailable` it is unusable so marking as exhausted
+                // Rollover is a no-op, so skipping to next trancheIn
+                i++;
+                continue;
+            }
+
+            // Perform rollover
+            _checkAndApproveMax(trancheIn, address(perp_), trancheInAmtAvailable);
```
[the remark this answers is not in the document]
aalavandhan: Using `rd.trancheInAmt` here results in some dust

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: I wanted to keep the interface generic so this could be a standard for tokenzied vaults. The rollover vault just accepts AMPL, however other vaults could accept different tokens.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,525 @@
...
+            _syncDeployedAsset(td.tranches[i]);
+        }
+        _syncAsset(underlying);
+    }
+
+    /// @dev Rolls over freshly tranched tokens from the given bond for older tranches (close to maturity) from perp.
+    /// @return The amount of perps rolled over.
+    function _rollover(IPerpetualTranche perp_, IBondController depositBond) private returns (uint256) {
+        // Tranches to be rolled over
+        TrancheData memory td = depositBond.getTrancheData();
```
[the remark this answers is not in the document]
aalavandhan: done

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,516 @@
...
+        for (uint256 i = 0; i < _deployed.length(); i++) {
+            ITranche tranche = deployedAt(i);
+            IBondController bond = IBondController(tranche.bond());
+
+            // if bond has mature, redeem the tranche token
+            if (bond.timeToMaturity() == 0) {
+                if (!bond.isMature()) {
+                    bond.mature();
+                }
+                bond.redeemMature(address(tranche), tranche.balanceOf(address(this)));
```
aalavandhan, line 438: Something to be cognizant of is that this can revert when the entire supply is redeemed and brick our `recoverAndRedeploy` workflow.

```
https://github.com/buttonwood-protocol/tranche/blob/main/contracts/BondController.sol#L225
```
Options:
1) We could wrap around try-catch
2) Add logic to compute amounts discounting the min condition
3) Handle manually if we ever run into this (deposit small amount of funds into the bond)

spot-contracts/contracts/vaults/RolloverVault.sol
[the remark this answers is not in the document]
nms-7: Makes sense 👍

spot-contracts/contracts/vaults/RolloverVault.sol
nms-7: Oh, that's right, good point. Could happen on `deposit` too if we don't deploy enough. I would say (2) compute the amounts, but since `MINIMUM_VALID_DEBT` isn't exposed, (1) try-catch works

spot-contracts/contracts/_utils/BondHelpers.sol
```diff
@@ -191,27 +191,33 @@ library BondHelpers {
         return (td, collateralBalances, trancheSupplies);
     }
 
-    /// @notice Given a bond, retrieves the collateral redeemable for
-    ///         each tranche held by the given address.
+    /// @notice Given a bond and a user address computes the tranche amounts proportional to the tranche ratios,
```
brandoniles, line 194: How about
```
For a given bond and user address, computes the maximum number of each of the bond's tranches the user is able to redeem before the bond's maturity. These tranche amounts necessarily match the bond's tranche ratios.
```

spot-contracts/contracts/_utils/BondHelpers.sol
```diff
@@ -191,27 +191,33 @@ library BondHelpers {
...
+            if (d < min) {
+                min = d;
+            }
+        }
 
-        uint256[] memory balances = new uint256[](td.trancheCount);
-        for (uint8 i = 0; i < td.trancheCount; i++) {
-            balances[i] = (td.tranches[i].balanceOf(u) * collateralBalances[i]) / trancheSupplies[i];
+        for (i = 0; i < td.trancheCount; i++) {
+            redeemableAmts[i] = (td.trancheRatios[i] * min) / TRANCHE_RATIO_GRANULARITY;
```
brandoniles, line 218: I'm confused by this section. I thought it would be the balances of tranches that can be redeemed. But I don't see the tranche balances of the users referenced anywhere here.

Also, wouldn't there be a different scalar for each tranche?

Say there's a bond that's configured as [20, 30, 50]
User's balances are [20, 25, 5000], so the limiting factor is tranche[1]
So it should return [16, 25, 40]

I would expect that we would need a different scalar number for each tranche balances. These would be 1x for the limiter and <= 1 for everything else. But it looks like this uses the same scalar for all of them.

spot-contracts/contracts/_utils/BondHelpers.sol
```diff
@@ -191,27 +191,33 @@ library BondHelpers {
...
+        uint256[] memory redeemableAmts = new uint256[](td.trancheCount);
+
+        // We calculate the minimum value of {trancheBal/trancheRatio} across tranches
+        uint256 min = type(uint256).max;
+        uint8 i;
+        for (i = 0; i < td.trancheCount && min != 0; i++) {
+            uint256 d = (td.tranches[i].balanceOf(u) * TRANCHE_RATIO_GRANULARITY) / td.trancheRatios[i];
+            if (d < min) {
+                min = d;
+            }
```
brandoniles, line 214: Would it be more efficient to break the min check into a dedicated condition in the loop?
```
if (min == 0) {
  return (td, redeemableAmts);
}
```
Short-circuits, but only if there's a zero balance so depends on the expected context.

spot-contracts/contracts/_utils/BondHelpers.sol
```diff
@@ -191,27 +191,33 @@ library BondHelpers {
...
-
-        (td, collateralBalances, trancheSupplies) = getTrancheCollateralizations(b);
+        TrancheData memory td = getTrancheData(b);
+        uint256[] memory redeemableAmts = new uint256[](td.trancheCount);
+
+        // We calculate the minimum value of {trancheBal/trancheRatio} across tranches
+        uint256 min = type(uint256).max;
+        uint8 i;
+        for (i = 0; i < td.trancheCount && min != 0; i++) {
+            uint256 d = (td.tranches[i].balanceOf(u) * TRANCHE_RATIO_GRANULARITY) / td.trancheRatios[i];
```
brandoniles, line 211: Use mulDiv?

brandoniles (commented):
Covers everything except the main vault file. I'll do that next

spot-contracts/contracts/_utils/BondHelpers.sol
aalavandhan: sure 👍

spot-contracts/contracts/_utils/BondHelpers.sol
aalavandhan: maybe the variables can be named differently

Consider A,B,Z with 0.2,0.3 and 0.5 and balances [10, 50, 20]

step one: we calculate balance/ratio -> 10/0.2, 50/0.3, 20/0.5 -> 50, 166.66, 40
step two: chose min (d=40)
step three: get amounts 40*0.2, 40*0.3, 40*0.5 -> [8, 12, 20]

d is essentially the maximum AMPL we can "re-tranche" from the existing balances

brandoniles: Ah ok,  yeah that makes sense. Code comments could clear that further. What about:

At top of loop-- ```Calculate how many underlying assets could be redeemed from each tranche balance, assuming other tranches are not an issue, and record the smallest amount.```

Then maybe rename `min` -> `minUnderlyingOut`

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,516 @@
...
+
+    /// @dev The set of the intermediate ERC-20 tokens when the underlying asset has been put to use.
+    ///      In the cause of this vault, they represent the tranche tokens held till maturity.
+    EnumerableSetUpgradeable.AddressSet private _deployed;
+
+    //-------------------------------------------------------------------------
+    // Data
+
+    /// @notice The perpetual token on which rollovers are performed.
+    IPerpetualTranche public perp;
```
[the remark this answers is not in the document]
aalavandhan: Right. I'm fine with keeping the v1 implementation specific. We can generalize down the road if we see perp evolve ..

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,516 @@
...
+    /// @param symbol ERC-20 Symbol of the vault token.
+    /// @param underlying_ ERC-20 address of the underlying token.
+    /// @param perp_ ERC-20 address of the perpetual tranche rolled over.
+    function init(
+        string memory name,
+        string memory symbol,
+        IERC20Upgradeable underlying_,
+        IPerpetualTranche perp_
+    ) public initializer {
+        __ERC20_init(name, symbol);
```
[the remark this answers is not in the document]
aalavandhan: Make sense adding it. Fine leaving it uncommented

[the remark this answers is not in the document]
aalavandhan: Sure will add it here. I dont think we added this to perp, might make sense to with this update (though it shouldn't affect the existing deployment, just hygiene).

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: Sure will do this in a later PR. Adding a todo

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,516 @@
...
+            _syncDeployedAsset(td.tranches[i]);
+        }
+        _syncAsset(underlying);
+
+        return td;
+    }
+
+    /// @dev Rolls over freshly tranched tokens from the given bond for older tranches (close to maturity) from perp.
+    /// @return The amount of perps rolled over.
+    function _rollover(IPerpetualTranche perp_, TrancheData memory td) private returns (uint256) {
```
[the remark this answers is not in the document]
aalavandhan: We could, however the `getTrancheData` call is expensive and its cheaper to pass td to `_rollover` than query the storage again. Agree it would not be part of the vault interface

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,516 @@
...
+
+    /// @notice Deploys deposited funds.
+    function deploy() public nonReentrant whenNotPaused {
+        TrancheData memory td = _tranche(perp.getDepositBond());
+        if (_rollover(perp, td) == 0) {
+            revert NoDeployment();
+        }
+    }
+
+    /// @notice Recovers deployed funds.
```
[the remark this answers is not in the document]
aalavandhan: It does not. With gelato we'll use `recoverAndRedeploy` which reverts if no assets were deployed.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,516 @@
...
+    // External & Public write methods
+
+    /// @notice Recovers deployed funds and redeploys them.
+    /// @dev Simply batches the `recover` and `deploy` functions.
+    function recoverAndRedeploy() external whenNotPaused {
+        recover();
+        deploy();
+    }
+
+    /// @notice Deploys deposited funds.
```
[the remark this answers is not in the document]
aalavandhan: yea will add a comment about reversion

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,516 @@
...
+            revert NoDeployment();
+        }
+    }
+
+    /// @notice Recovers deployed funds.
+    function recover() public nonReentrant whenNotPaused {
+        _redeemTranches();
+    }
+
+    /// @notice Deposits the provided asset from {msg.sender} into the vault and mints notes.
```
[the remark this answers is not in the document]
aalavandhan: 👍

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,516 @@
...
+        uint256 notes = (totalSupply_ > 0) ? amount.mulDiv(totalSupply_, getTVL()) : (amount * INITIAL_RATE);
+
+        underlying.safeTransferFrom(_msgSender(), address(this), amount);
+        _syncAsset(underlying);
+
+        _mint(_msgSender(), notes);
+        return notes;
+    }
+
+    struct RedemptionPreview {
```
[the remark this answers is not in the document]
aalavandhan: 👍

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+
+        // burn notes
+        _burn(_msgSender(), notes);
+
+        // aggregating vault assets to be redeemed
+        redemptions[0].token = underlying;
+        for (uint256 i = 0; i < deployedCount_; i++) {
+            redemptions[i + 1].token = deployedAt(i);
+        }
+        for (uint256 i = 0; i < earnedCount_; i++) {
```
brandoniles, line 239: Think we really need this `for` loop, and the `earnedCount` function?

I think we already see this as not being able to configure the earned token after deployment, so I think we know it will always be 1 for this instance.

We assume there's 1 underlying in line 227, even though the deposit function allows you to specify which token to deposit. So we're already straddling two level of generalization. I'd just go with the simplest here.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+    //
+    // In the case of this vault, the "earned" assets are the perp tokens themselves.
+    // The reward (or yield) for performing rollovers is paid out in perp tokens.
+
+    /// @notice The ERC20 token that can be deposited into this vault.
+    IERC20Upgradeable public underlying;
+
+    /// @dev The set of the intermediate ERC-20 tokens when the underlying asset has been put to use.
+    ///      In the case of this vault, they represent the tranche tokens held before maturity.
+    EnumerableSetUpgradeable.AddressSet private _deployed;
```
brandoniles, line 109: Any reason to make this private? We could remove the `deployedAt` function that's down below.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+
+        TokenAmount[] memory redemptions = new TokenAmount[](assetCount);
+
+        // burn notes
+        _burn(_msgSender(), notes);
+
+        // aggregating vault assets to be redeemed
+        redemptions[0].token = underlying;
+        for (uint256 i = 0; i < deployedCount_; i++) {
+            redemptions[i + 1].token = deployedAt(i);
```
brandoniles, line 237: Could be safer to use `redemptions.push()`, instead of doing index math.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+        for (uint256 i = 0; i < assetCount; i++) {
+            redemptions[i].amount = _calculateAssetShare(redemptions[i].token, notes, totalNotes);
+            redemptions[i].token.safeTransfer(_msgSender(), redemptions[i].amount);
+            _syncAsset(redemptions[i].token);
+        }
+
+        return redemptions;
+    }
+
+    /// @notice Total value of assets currently held by the vault denominated in the underlying asset.
```
brandoniles, line 253: I'd change `@notice` to `@return`

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+        return totalAssets;
+    }
+
+    //--------------------------------------------------------------------------
+    // External & Public read methods
+
+    /// @notice Fetches the vault's asset token balance.
+    /// @param token The address of the asset ERC-20 token held by the vault.
+    /// @return The token balance.
+    function assetBalance(IERC20Upgradeable token) external view returns (uint256) {
```
brandoniles, line 289: I'd name this `vaultAssetBalance`.

This function assumes it's a vault asset, but `transferERC20()` reverts when it isn't. So the names should be really clear in each case to avoid confusion.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+        return isVaultAsset(token) ? token.balanceOf(address(this)) : 0;
+    }
+
+    /// @notice Checks if the given token is held by the vault.
+    /// @param token The address of a token to check.
+    function isVaultAsset(IERC20Upgradeable token) public view returns (bool) {
+        return token == underlying || _deployed.contains(address(token)) || address(perp) == address(token);
```
brandoniles, line 296: This is an instance where it assumes there's one earned token, but other parts of the code doesn't. I think let's just assume there's one in rest of code too.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+    }
+
+    /// @notice Total count of deployed asset tokens held by the vault.
+    /// @return The count.
+    function deployedCount() public view returns (uint256) {
```
brandoniles, line 301: I think we can make this `external`, then just call `.length()` in our own code

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: Makes sense 👍

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: I dont think, making `AddressSet` public doesn't automatically give us the public getter methods. They have to be implemented ..

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+        /// @notice The amount redeemed.
+        uint256 amount;
+    }
+
+    /// @notice Burns notes and sends a proportional share vault's assets back to {msg.sender}.
+    /// @param notes The amount of notes to be burnt.
+    /// @return The list of asset tokens and amounts redeemed.
+    function redeem(uint256 notes) external nonReentrant whenNotPaused returns (TokenAmount[] memory) {
+        uint256 totalNotes = totalSupply();
+        uint256 deployedCount_ = deployedCount();
```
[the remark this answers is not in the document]
aalavandhan: 👍

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+            redemptions[i].amount = _calculateAssetShare(redemptions[i].token, notes, totalNotes);
+            redemptions[i].token.safeTransfer(_msgSender(), redemptions[i].amount);
+            _syncAsset(redemptions[i].token);
+        }
+
+        return redemptions;
+    }
+
+    /// @notice Total value of assets currently held by the vault denominated in the underlying asset.
+    function getTVL() public returns (uint256) {
```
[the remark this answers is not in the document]
aalavandhan: The `getAvgPrice` call within this function isn't a view function as it calls `updateState`

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+        uint256 totalAssets = 0;
+
+        // The underlying balance
+        totalAssets += underlying.balanceOf(address(this));
+
+        // The deployed asset value denominated in the underlying
+        for (uint256 i = 0; i < _deployed.length(); i++) {
+            ITranche tranche = deployedAt(i);
```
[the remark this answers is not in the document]
aalavandhan: done 👍

spot-contracts/contracts/vaults/RolloverVault.sol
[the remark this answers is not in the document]
aalavandhan: I like keeping it part of the standard interface of off-chain calls if anything

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: Yea agree, lets assume one earned token for now everywhere

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: I dont think push is available for memory arrays

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+        _redeemTranches();
+    }
+
+    /// @notice Deposits the provided asset from {msg.sender} into the vault and mints notes.
+    /// @param token The address of the asset to deposit.
+    /// @param amount The amount tokens to be deposited into the vault.
+    /// @return The amount of notes.
+    function deposit(IERC20Upgradeable token, uint256 amount) external nonReentrant whenNotPaused returns (uint256) {
+        // NOTE: The vault only accepts the underlying asset tokens.
+        if (token != underlying) {
```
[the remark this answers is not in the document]
aalavandhan: I dont think we can have constructors in the proxy?

We could still get rid of the `token` parameter, don't think we need to generalize now. I'm okay with `underlying` not being strictly immutable, if the code doesn't have methods change it.

https://forum.openzeppelin.com/t/upgradable-contracts-instantiating-an-immutable-value/28763/2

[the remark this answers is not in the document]
brandoniles: Agree, that all sounds good to me 👍

spot-contracts/contracts/vaults/RolloverVault.sol
brandoniles: Ah of course, was thinking about built-in arrays.

spot-contracts/contracts/vaults/RolloverVault.sol
brandoniles: Weird, yeah there are no dynamic memory arrays...

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -217,30 +217,26 @@ contract RolloverVault is
...
     /// @return The list of asset tokens and amounts redeemed.
     function redeem(uint256 notes) external nonReentrant whenNotPaused returns (TokenAmount[] memory) {
         uint256 totalNotes = totalSupply();
-        uint256 deployedCount_ = deployedCount();
-        uint256 earnedCount_ = earnedCount();
-        uint256 assetCount = 1 + deployedCount_ + earnedCount_;
-
-        TokenAmount[] memory redemptions = new TokenAmount[](assetCount);
+        uint256 deployedCount_ = _deployed.length();
+        uint256 assetCount = 2 + deployedCount_;
```
brandoniles, line 227: I could see a class of bugs open up if `underlying` and `earned` are the same token.

I don't see this being the case for this deployment, so perhaps the easiest way to handle it is to add an assert statement in initialization somewhere that they're not equal.

It might work as is... but it's weird to have an asset appear twice in a list of assets, or the number of assets not equal the length of the list of assets, etc. It's a case I'd rather not have to test for or think about.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -217,30 +217,26 @@ contract RolloverVault is
...
+        TokenAmount[] memory redemptions = new TokenAmount[](assetCount);
         redemptions[0].token = underlying;
         for (uint256 i = 0; i < deployedCount_; i++) {
-            redemptions[i + 1].token = deployedAt(i);
-        }
-        for (uint256 i = 0; i < earnedCount_; i++) {
-            redemptions[i + 1 + deployedCount_].token = earnedAt(i);
+            redemptions[i + 1].token = IERC20Upgradeable(_deployed.at(i));
         }
+        redemptions[deployedCount_ + 1].token = IERC20Upgradeable(perp);
```
brandoniles, line 235: Oh, ok--we're assuming that the earned token is perp. Ties into the comment above, because presumably perp and underlying can't be the same.

Think it's cleaner to take the earned token as an initialization parameter, then not worry about it?

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -283,46 +276,43 @@ contract RolloverVault is
...
+    /// @return The token address from the deployed asset token list by index.
+    function deployedAt(uint256 i) external view returns (IERC20Upgradeable) {
+        return IERC20Upgradeable(_deployed.at(i));
     }
 
-    /// @notice Total count of earned income tokens held by the vault.
-    /// @return The count.
-    function earnedCount() public pure returns (uint256) {
+    /// @return Total count of earned income tokens held by the vault.
+    function earnedCount() external pure returns (uint256) {
```
brandoniles, line 298: Keeping these because you see them being part of the generalized vault interface?

brandoniles (commented):
_rollover

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,508 @@
...
+
+        // Batch rollover:
+        // For each tranche token (held by the vault) to be rolled over,
+        // and each tranche token up for rollover (held by perp),
+        // Execute rollover and move on to the next pair.
+        uint256 totalPerpRolledOver = 0;
+        uint8 i = 0;
+        uint256 j = 0;
+        while (i < td.trancheCount && j < rolloverTokens.length) {
+            ITranche trancheIn = td.tranches[i];
```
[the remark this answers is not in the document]
aalavandhan: In and out refer to going into and out of perp

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: We can assert `underlying != earned` in the initializer

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,508 @@
...
+        IERC20Upgradeable[] memory rolloverTokens = perp_.getReserveTokensUpForRollover();
+
+        // Batch rollover:
+        // For each tranche token (held by the vault) to be rolled over,
+        // and each tranche token up for rollover (held by perp),
+        // Execute rollover and move on to the next pair.
+        uint256 totalPerpRolledOver = 0;
+        uint8 i = 0;
+        uint256 j = 0;
+        while (i < td.trancheCount && j < rolloverTokens.length) {
```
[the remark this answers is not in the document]
aalavandhan: okay breaking into two loops

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: yea we'll need it when we add a secondary perp

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: The question if we keep track of "earned" and infer "perp", or keep track of perp and infer earned. I went back and forth and chose the latter ..

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,508 @@
...
+        // NOTE: The first element of the list is the mature tranche,
+        //       there after the list is NOT ordered by maturity.
+        IERC20Upgradeable[] memory rolloverTokens = perp_.getReserveTokensUpForRollover();
+
+        // Batch rollover:
+        // For each tranche token (held by the vault) to be rolled over,
+        // and each tranche token up for rollover (held by perp),
+        // Execute rollover and move on to the next pair.
+        uint256 totalPerpRolledOver = 0;
+        uint8 i = 0;
```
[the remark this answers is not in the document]
aalavandhan: 👍

spot-contracts/contracts/vaults/RolloverVault.sol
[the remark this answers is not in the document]
aalavandhan: Two loops does not apply here right?

We have two sets of tokens => (a,b,c) (1,2,3,4), each with individual balances

What we need to do is pick a token from the first set say a, then pick a token from the 2nd set say 1. Pair them to execute a rollover(a,1). After this operation, either a might remain or 1 might remain or neither. 

In the case a remains, we pair rollover(a,2)
In the case 1 remains, we pair rollover(b,1)
In the case none remains, we pair rollover(b,2)

Similarly we continue till we reach the end of either of the list and exhaust the balances.

We aren't pairing every member of the two sets with each other, which is what will happen in a nested loop.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,517 @@
...
+    function redeem(uint256 notes) external nonReentrant whenNotPaused returns (TokenAmount[] memory) {
+        uint256 totalNotes = totalSupply();
+        uint256 deployedCount_ = deployedCount();
+        uint256 earnedCount_ = earnedCount();
+        uint256 assetCount = 1 + deployedCount_ + earnedCount_;
+
+        TokenAmount[] memory redemptions = new TokenAmount[](assetCount);
+
+        // burn notes
+        _burn(_msgSender(), notes);
```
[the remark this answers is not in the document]
aalavandhan: done

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: ah got it. But at the storage level we just keep track of `perp` and the internal contract functions point to `perp` when they want to access `earned`. (which is how it currently works)

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,513 @@
...
+        if (balance == 0 && isHeld) {
+            // Removes token into the deployed assets list.
+            _deployed.remove(address(token));
+        }
+
+        return balance;
+    }
+
+    /// @dev Approves the spender to spend an infinite tokens from the vault's balance.
+    // NOTE: Only audited & immutable spender contracts should have infinite approvals.
```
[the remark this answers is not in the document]
aalavandhan: In general I think infinite approvals are fine for vetted tokens, like ampl, tranche and perp. So I'm fine with doing an infinite approval here as these will be the tokens that the vault holds. (This should give us significant gas savings)

If we want to be conservative (as the vault will custody a lot of funds), we can just approve the necessary amount and pay the extra gas cost.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -0,0 +1,513 @@
...
+    function _syncDeployedAsset(IERC20Upgradeable token) private returns (uint256) {
+        uint256 balance = _syncAsset(token);
+        bool isHeld = _deployed.contains(address(token));
+
+        if (balance > 0 && !isHeld) {
+            // Inserts new token into the deployed assets list.
+            _deployed.add(address(token));
+        }
+
+        if (balance == 0 && isHeld) {
```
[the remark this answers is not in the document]
aalavandhan: Right once they reach maturity the should be removed. Additionally we could add a check which limits the size of the `_deployed` list.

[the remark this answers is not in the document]
brandoniles: Great, I think it's fine as is 👍

spot-contracts/contracts/vaults/RolloverVault.sol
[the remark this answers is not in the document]
brandoniles: Yep makes sense. The new variable names and comments makes everything a lot easier to understand, thanks!

spot-contracts/contracts/vaults/RolloverVault.sol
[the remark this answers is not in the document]
brandoniles: With the amount of funds potentially at stake in the future, I do like the idea of approving only as much as is needed. AMPL is very stable, and hopefully tranches are as well, but tranche has already been updated once. If a future update contained a bug, it would be good to minimize collateral damage as much as possible.

Not to mention if someone were able to exploit some other weakness in the vault to inject a custom malicious token into the assets list.

spot-contracts/contracts/vaults/RolloverVault.sol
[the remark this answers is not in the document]
brandoniles: Right before the loop on line 244?
```
// transferring assets out proportionally
        for (uint256 i = 0; i < assetCount; i++) {
```

spot-contracts/contracts/vaults/RolloverVault.sol
brandoniles: Ah right, nvm

spot-contracts/contracts/vaults/RolloverVault.sol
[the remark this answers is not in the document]
aalavandhan: 👍

brandoniles (approved):
LGTM

PR #115, opened 2022-11-29
Rollover vault tests

aalavandhan:
* Smock for contract mocks
* Added tests for rollover vault flows


spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -454,8 +459,8 @@ contract RolloverVault is
         // NOTE: We traverse the deployed set in the reverse order
         //       as deletions involve swapping the deleted element to the
         //       end of the set and removing the last element.
-        for (uint256 i = _deployed.length() - 1; i >= 0; i--) {
```
aalavandhan, line 457: i-- underflows when i = 0

brandoniles: Nice catch. I still do like the original style of
```
(uint256 i = _deployed.length() - 1; i >= 0; i--) {
```
since it removes the possibility of forgetting to decrement inside the loop on each use. The new implementation might cost more gas as well, since there are more decrements.

aalavandhan: This will underflow anyway right? we cant have `i--` and `i>=0`. when i becomes 0, and the loop tries to decrement it will error out 
```
(uint256 i = _deployed.length() - 1; i >= 0; i--) {
```

brandoniles: Ah of course... `i >= 0` literally can't be false when i is a uint, reverts not even withstanding. My bad

brandoniles (approved):
LGTM

PR #128, opened 2023-01-05
Issuer update v1.1.0

aalavandhan:
- updated issuer book keeping to keep track of active bond indices
- Added helper to execute `mature` on active bonds when they are up for maturity

### DIFF
--- a/spot-contracts/contracts/BondIssuer.sol
+++ b/spot-contracts/contracts/BondIssuer.sol
@@ -4,9 +4,14 @@ pragma solidity ^0.8.17;
 import { OwnableUpgradeable } from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 import { EnumerableSetUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/structs/EnumerableSetUpgradeable.sol";
 
+import { BondHelpers } from "./_utils/BondHelpers.sol";
+
 import { IBondFactory } from "./_interfaces/buttonwood/IBondFactory.sol";
 import { IBondController } from "./_interfaces/buttonwood/IBondController.sol";
 import { IBondIssuer } from "./_interfaces/IBondIssuer.sol";
+
+/// @notice Expected at least one mature bond.
+error NoMaturedBonds();
 
 /**
  *  @title BondIssuer
@@ -18,6 +23,7 @@ import { IBondIssuer } from "./_interfaces/IBondIssuer.sol";
  */
 contract BondIssuer is IBondIssuer, OwnableUpgradeable {
     using EnumerableSetUpgradeable for EnumerableSetUpgradeable.AddressSet;
+    using BondHelpers for IBondController;
 
     /// @dev Using the same granularity as the underlying buttonwood tranche contracts.
     ///      https://github.com/buttonwood-protocol/tranche/blob/main/contracts/BondController.sol
@@ -50,10 +56,15 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
     ///      then the issue window opens at Friday 2AM GMT every week.
     uint256 public issueWindowOffsetSec;
 
-    /// @notice An enumerable list to keep track of bonds issued by this issuer.
-    /// @dev Bonds are only added and never removed, thus the last item will always point
-    ///      to the latest bond.
-    EnumerableSetUpgradeable.AddressSet private _issuedBonds;
+    /// @dev List of all active bonds, i.e) issued bonds which have not been matured.
+    EnumerableSetUpgradeable.AddressSet private _activeBonds;
+
+    /// @dev List of "inactive", matured bonds.
+    ///      Once a bond is matured it's moved from the `active` list into this one.
+    EnumerableSetUpgradeable.AddressSet private _maturedBonds;
+
+    /// @dev The reference to the most recently issued bond.
+    IBondController private _lastBond;
 
     /// @notice The timestamp when the issue window opened during the last issue.
     uint256 public lastIssueWindowTimestamp;
@@ -113,7 +124,35 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
 
     /// @inheritdoc IBondIssuer
     function isInstance(IBondController bond) external view override returns (bool) {
-        return _issuedBonds.contains(address(bond));
+        return (_activeBonds.contains(address(bond)) || _maturedBonds.contains(address(bond)));
+    }
+
+    /// @inheritdoc IBondIssuer
+    /// @dev Reverts if none of the active bonds are mature.
+    function matureAll() external {
+        bool bondsMature = false;
+        uint256 activeCount_ = _activeBonds.length();
+
+        // NOTE: We traverse the active list in the reverse order as deletions involve
+        //       swapping the deleted element to the end of the list and removing the last element.
+        for (uint256 i = activeCount_; i > 0; i--) {
+            IBondController bond = IBondController(_activeBonds.at(i - 1));
+            if (bond.timeToMaturity() <= 0) {
+                if (!bond.isMature()) {
+                    bond.mature();
+                }
+
+                _activeBonds.remove(address(bond));
+                _maturedBonds.add(address(bond));
+                emit BondMature(bond);
+
+                bondsMature = true;
+            }
+        }
+
+        if (!bondsMature) {
+            revert NoMaturedBonds();
+        }
     }
 
     /// @inheritdoc IBondIssuer
@@ -131,7 +170,8 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
             bondFactory.createBond(collateral, trancheRatios, lastIssueWindowTimestamp + maxMaturityDuration)
         );
 
-        _issuedBonds.add(address(bond));
+        _activeBonds.add(address(bond));
+        _lastBond = bond;
 
         emit BondIssued(bond);
     }
@@ -140,17 +180,29 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
     /// @dev Lazily issues a new bond when the time is right.
     function getLatestBond() external override returns (IBondController) {
         issue();
-        // NOTE: The latest bond will be at the end of the list.
-        return IBondController(_issuedBonds.at(_issuedBonds.length() - 1));
+        return _lastBond;
     }
 
     /// @inheritdoc IBondIssuer
     function issuedCount() external view override returns (uint256) {
-        return _issuedBonds.length();
+        return _activeBonds.length() + _maturedBonds.length();
     }
 
     /// @inheritdoc IBondIssuer
+    /// @dev Internally, we first iterate through the active list and then move on to the mature list.
+    ///      This list is NOT guaranteed to be ordered.
     function issuedBondAt(uint256 index) external view override returns (IBondController) {
-        return IBondController(_issuedBonds.at(index));
+        uint256 activeCount_ = _activeBonds.length();
+        if (index < activeCount_) {
+            return IBondController(_activeBonds.at(index));
+        } else {
+            return IBondController(_maturedBonds.at(index - activeCount_));
+        }
+    }
+
+    /// @notice Returns the count of issued bonds which are still active (i.e have not matured).
+    /// @return The number of active bonds.
+    function activeCount() external view returns (uint256) {
+        return _activeBonds.length();
     }
 }
--- a/spot-contracts/contracts/BondIssuer.sol
+++ b/spot-contracts/contracts/BondIssuer.sol
@@ -8,10 +8,7 @@ import { BondHelpers } from "./_utils/BondHelpers.sol";
 
 import { IBondFactory } from "./_interfaces/buttonwood/IBondFactory.sol";
 import { IBondController } from "./_interfaces/buttonwood/IBondController.sol";
-import { IBondIssuer } from "./_interfaces/IBondIssuer.sol";
-
-/// @notice Expected at least one mature bond.
-error NoMaturedBonds();
+import { IBondIssuer, NoMaturedBonds } from "./_interfaces/IBondIssuer.sol";
 
 /**
  *  @title BondIssuer
@@ -127,7 +124,7 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
 
     /// @inheritdoc IBondIssuer
     /// @dev Reverts if none of the active bonds are mature.
-    function matureAll() external {
+    function matureActive() external {
         bool bondsMature = false;
 
         // NOTE: We traverse the active index list in the reverse order as deletions involve
--- a/spot-contracts/contracts/BondIssuer.sol
+++ b/spot-contracts/contracts/BondIssuer.sol
@@ -187,4 +187,15 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
     function issuedBondAt(uint256 index) external view override returns (IBondController) {
         return IBondController(_issuedBonds.at(index));
     }
+
+    /// @inheritdoc IBondIssuer
+    function activeCount() external view override returns (uint256) {
+        return _activeBondIDXs.length;
+    }
+
+    /// @inheritdoc IBondIssuer
+    /// @dev This is NOT ordered by issuance time.
+    function activeBondAt(uint256 index) external view override returns (IBondController) {
+        return IBondController(_issuedBonds.at(_activeBondIDXs[index]));
+    }
 }
--- a/spot-contracts/contracts/BondIssuer.sol
+++ b/spot-contracts/contracts/BondIssuer.sol
@@ -188,14 +188,17 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
         return IBondController(_issuedBonds.at(index));
     }
 
-    /// @inheritdoc IBondIssuer
-    function activeCount() external view override returns (uint256) {
+    /// @notice Number of bonds issued by this issuer which have not yet reached their maturity date.
+    /// @return Number of active bonds.
+    function activeCount() external view returns (uint256) {
         return _activeBondIDXs.length;
     }
 
-    /// @inheritdoc IBondIssuer
+    /// @notice The bond address from the active list by index.
     /// @dev This is NOT ordered by issuance time.
-    function activeBondAt(uint256 index) external view override returns (IBondController) {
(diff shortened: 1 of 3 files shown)


spot-contracts/contracts/BondIssuer.sol
```diff
@@ -50,10 +53,15 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
     ///      then the issue window opens at Friday 2AM GMT every week.
     uint256 public issueWindowOffsetSec;
 
-    /// @notice An enumerable list to keep track of bonds issued by this issuer.
-    /// @dev Bonds are only added and never removed, thus the last item will always point
-    ///      to the latest bond.
-    EnumerableSetUpgradeable.AddressSet private _issuedBonds;
```
aalavandhan, line 56: The issuer is not upgradable, we will deploy a new one. Its fine to modify the storage layout.

spot-contracts/contracts/BondIssuer.sol
```diff
@@ -50,10 +56,15 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
...
+    /// @dev An list of all active bonds, i.e) bonds which have NOT mature.
```
brandoniles, line 59:
```suggestion
    /// @dev List of all active bonds, i.e) issued bonds which have not been matured.
```

spot-contracts/contracts/BondIssuer.sol
```diff
@@ -50,10 +56,15 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
...
+    EnumerableSetUpgradeable.AddressSet private _activeBonds;
+
+    /// @dev An list of "inactive" or mature bonds.
```
brandoniles, line 62:
```suggestion
    /// @dev List of "inactive", matured bonds.
```

spot-contracts/contracts/BondIssuer.sol
```diff
@@ -50,10 +56,15 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
...
+    ///      Once a bond matures its moved from the `active` list into this one.
```
brandoniles, line 63:
```suggestion
    ///      Once a bond is matured it's moved from the `active` list into this one.
```

spot-contracts/contracts/BondIssuer.sol
```diff
@@ -50,10 +56,15 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
...
+    EnumerableSetUpgradeable.AddressSet private _matureBonds;
```
brandoniles, line 64: Apologies this might seem really minor, but we could rename this to `_maturedBonds`.

This makes it clear that the bonds here have explicitly "been matured", meaning that mature() has been executed.
I make a distinction between this state and the state where a bond has reached the maturity time, but hasn't yet been matured in contract storage.

spot-contracts/contracts/BondIssuer.sol
```diff
@@ -113,7 +124,35 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
...
+
+    /// @inheritdoc IBondIssuer
+    /// @dev Reverts if none of the active bonds are mature.
+    function matureAll() external {
+        bool bondsMature = false;
+        uint256 activeCount_ = _activeBonds.length();
+
+        // NOTE: We traverse the active list in the reverse order as deletions involve
+        //       swapping the deleted element to the end of the list and removing the last element.
+        for (uint256 i = activeCount_; i > 0; i--) {
```
brandoniles, line 138: Shouldn't this be...
```
for (uint256 i = activeCount_ - 1; i >= 0; i--) {
```

Edit: Ah, I see the subtraction on the next line.

A more idiomatic style is to cycle over the range in the storage container's index space, then use that normally in the code that follows.

This has a couple benefits:
- It reduces the chance of future bugs where someone uses `i` again but forgets to decrement it on each use.
-  If `i` is used more than once, it saves a small amount of gas.
- It's idiomatic, so future readers won't have their off-by-one alerts when they read it and it's easier to read quickly.

spot-contracts/contracts/BondIssuer.sol
```diff
@@ -113,7 +124,35 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
...
     function isInstance(IBondController bond) external view override returns (bool) {
-        return _issuedBonds.contains(address(bond));
+        return (_activeBonds.contains(address(bond)) || _matureBonds.contains(address(bond)));
+    }
+
+    /// @inheritdoc IBondIssuer
+    /// @dev Reverts if none of the active bonds are mature.
+    function matureAll() external {
+        bool bondsMature = false;
+        uint256 activeCount_ = _activeBonds.length();
```
brandoniles, line 134: Do we use the trailing suffix consistently? Is it supposed to mean local scope?

spot-contracts/contracts/BondIssuer.sol
```diff
@@ -113,7 +124,35 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
...
+
+        // NOTE: We traverse the active list in the reverse order as deletions involve
+        //       swapping the deleted element to the end of the list and removing the last element.
+        for (uint256 i = activeCount_; i > 0; i--) {
+            IBondController bond = IBondController(_activeBonds.at(i - 1));
+            if (bond.timeToMaturity() == 0) {
```
brandoniles, line 140:
```suggestion
            if (bond.timeToMaturity() <= 0) {
```
Common defensive programming technique (unless there's a gas reason not to use it).

spot-contracts/contracts/BondIssuer.sol
```diff
@@ -113,7 +124,35 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
...
+
+                _activeBonds.remove(address(bond));
+                _matureBonds.add(address(bond));
+                emit BondMature(bond);
+
+                bondsMature = true;
+            }
+        }
+
+        if (!bondsMature) {
```
brandoniles, line 153: Is there a benefit to having this revert? It seems like we're using an error reporting feature for returning data to the caller.

In the case of no maturity events, I'd expect it to be a no-op, return a bool, or return an integer representing the number of bonds that matured (0).

spot-contracts/contracts/BondIssuer.sol
```diff
@@ -140,17 +180,29 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
     /// @dev Lazily issues a new bond when the time is right.
     function getLatestBond() external override returns (IBondController) {
         issue();
-        // NOTE: The latest bond will be at the end of the list.
-        return IBondController(_issuedBonds.at(_issuedBonds.length() - 1));
+        return _lastBond;
```
brandoniles, line 183: Is the purpose of adding the new contract storage variable to optimize this function call?

This adds more gas cost to the `issue()` function and therefore `updateState()` on Perp.

spot-contracts/contracts/BondIssuer.sol
```diff
@@ -140,17 +180,29 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
...
 
     /// @inheritdoc IBondIssuer
     function issuedCount() external view override returns (uint256) {
-        return _issuedBonds.length();
+        return _activeBonds.length() + _matureBonds.length();
     }
 
     /// @inheritdoc IBondIssuer
+    /// @dev Internally, we first iterate through the active list and then move on to the mature list.
+    ///      This list is NOT guaranteed to be ordered.
```
brandoniles, line 193: I guess more specifically:
- `[0, activeBonds.len)` will be ordered by increasing issued date,
-  `[activeBonds.len, numBonds)` will be ordered by increasing maturity date

spot-contracts/contracts/BondIssuer.sol
aalavandhan: Yea that makes sense 👍

spot-contracts/contracts/BondIssuer.sol
aalavandhan: We usually use the trailing `_` when there is a namespace conflict. There's already a method called `activeCount` and hence the variable is `activeCount_`.

The `_` when prepended to a variable indicates its private.

spot-contracts/contracts/BondIssuer.sol
aalavandhan: The reverts are useful when we hook this up to multiple keeper services like gelato, chainlink  and probably our own. The first successful call succeeds and the others fail, so the off-chain keeper script doesn't execute.

spot-contracts/contracts/BondIssuer.sol
aalavandhan: ^ This will throw an error when `activeCount_ = 0`

spot-contracts/contracts/BondIssuer.sol
aalavandhan: 👍

spot-contracts/contracts/BondIssuer.sol
aalavandhan: yea makes sense. Maybe we could just return the last item in `_activeBonds` ?

spot-contracts/contracts/BondIssuer.sol
aalavandhan: cool. Added this to the comment

spot-contracts/contracts/BondIssuer.sol
aalavandhan: We need to keep track of the last issued bond somehow. We can't rely on fetching the last element in `_activeBonds` as `_activeBonds` could be empty (if all the bonds mature).  `_maturedBonds` on the other hand will be ordered by maturity date and not by issue date (in practice though it should be ordered by issue date but its not guaranteed).

spot-contracts/contracts/BondIssuer.sol
aalavandhan: "[0, activeBonds.len) will be ordered by increasing issued date," -> I dont think this is true.

Since when a bond matures and is removed from the activeBonds list, the last element gets swapped in place messing up the ordering.

spot-contracts/contracts/BondIssuer.sol
aalavandhan: More generally I'm currently leaning toward poke-able functions (like rebase) reverting. That way off-chain callers can be dumb and just keep polling until its executable and trigger it.

spot-contracts/contracts/BondIssuer.sol
aalavandhan: This does add an an additional storage update on "issue" but it's a onetime cost.

spot-contracts/contracts/BondIssuer.sol
```diff
@@ -140,17 +180,30 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
...
     /// @inheritdoc IBondIssuer
     function issuedCount() external view override returns (uint256) {
-        return _issuedBonds.length();
+        return _activeBonds.length() + _maturedBonds.length();
     }
 
     /// @inheritdoc IBondIssuer
+    /// @dev Internally, we first iterate through the active list and then move on to the mature list.
+    ///      [0, activeBonds.len) is NOT guaranteed to be ordered,
+    ///      [activeBonds.len, numBonds) will be ordered by increasing maturity date.
```
nms-7, line 194: not sure `_maturedBonds` is guaranteed to be ordered by increasing maturity date, even if it likely will in practice. 

example: `_activeBonds`, is, by maturity, [D, D+1, D+2], but after a `matureAll() after D, but before D+1, becomes [D+2, D+1]. Then a new bond gets added so its, [D+2, D+1, D+3]. If `matureAll()` does not get called again until D+3, the tail of `_matureBonds` would be [D+3, D+1, D+2]. 

unless we add `matureAll()` calls in, e.g. `issue()`, the only guarantee is that the function is poked externally and regularly

aalavandhan: yea we should assume that its not ordered

spot-contracts/contracts/_interfaces/IBondIssuer.sol
```diff
@@ -8,10 +8,17 @@ interface IBondIssuer {
     /// @param bond The newly issued bond.
     event BondIssued(IBondController bond);
 
+    /// @notice Event emitted when a bond has matured.
+    /// @param bond The mature bond.
```
brandoniles, line 12:
```suggestion
    /// @param bond The matured bond.
```

spot-contracts/contracts/_interfaces/IBondIssuer.sol
```diff
@@ -8,10 +8,17 @@ interface IBondIssuer {
...
+    event BondMature(IBondController bond);
```
brandoniles, line 13:
```suggestion
    event BondMatured(IBondController bond);
```

spot-contracts/contracts/_interfaces/IBondIssuer.sol
```diff
@@ -8,10 +8,17 @@ interface IBondIssuer {
...
+
     /// @notice The address of the underlying collateral token to be used for issued bonds.
     /// @return Address of the collateral token.
     function collateral() external view returns (address);
 
+    /// @notice Invokes `mature` on issued bonds.
```
brandoniles, line 19: Should we comment that it fails if there are no mature operations to execute? This would be behavior that clients like Gelato would rely on.

One step further would be to name the function `matureAllOrRevert()`

spot-contracts/contracts/BondIssuer.sol
```diff
@@ -4,10 +4,15 @@ pragma solidity ^0.8.17;
 import { OwnableUpgradeable } from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 import { EnumerableSetUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/structs/EnumerableSetUpgradeable.sol";
 
+import { BondHelpers } from "./_utils/BondHelpers.sol";
```
brandoniles, line 7: Normally I expect interface imports to come before library imports. Maybe move this down after the interfaces.

spot-contracts/contracts/BondIssuer.sol
```diff
@@ -4,10 +4,15 @@ pragma solidity ^0.8.17;
...
+
 import { IBondFactory } from "./_interfaces/buttonwood/IBondFactory.sol";
 import { IBondController } from "./_interfaces/buttonwood/IBondController.sol";
 import { IBondIssuer } from "./_interfaces/IBondIssuer.sol";
 
+/// @notice Expected at least one mature bond.
+error NoMaturedBonds();
```
brandoniles, line 14: If we consider that reverting is part of the interface definition (see below), then this could go in the interface file instead of here.

brandoniles (commented):
Much cleaner 👍

spot-contracts/contracts/_interfaces/IBondIssuer.sol
aalavandhan: I don't want to change the existing interface which the other issuers on mainnet implement

spot-contracts/contracts/_interfaces/IBondIssuer.sol
aalavandhan: Seems like `matureActive()` is more appropriate. I think its reasonable to assume that it reverts if no active bonds exist. Theres a `@dev` comment on above the function implementation about the revert ..

spot-contracts/contracts/BondIssuer.sol
aalavandhan: Currently we import the libraries first and then the interfaces so for now will keep it consistent across the project

nms-7 (approved):

spot-contracts/contracts/_interfaces/IBondIssuer.sol
```diff
@@ -31,4 +41,12 @@ interface IBondIssuer {
     /// @notice The bond address from the issued list by index.
     /// @return Address of the bond.
     function issuedBondAt(uint256 index) external view returns (IBondController);
+
+    /// @notice The total number of bonds issued by this issuer which are currently active.
+    /// @return Number of active bonds.
+    function activeCount() external view returns (uint256);
+
+    /// @notice The bond address from the active list by index.
+    /// @return Address of the active bond.
```
brandoniles, line 50: This is actually a very weird interface definition, which makes me think our interface definition isn't quite right.

`index` is not the index of the bond in the bonds container. It's an index to an array that gives you an index into the bond container.

Also, `index` is an index into a datastructure that's not mentioned or documented anywhere else in the interface file. And the order of the indexes in this datastructure depends on the implementation, which isn't constrained here.

Does it need to be here? If so, does it make sense to have the activeBondIDX as part of the interface?

aalavandhan: The index of the bonds in the bonds container is private isn't exposed publicly. 

As far as the outside world is concerned, The there are `activeCount()` active bonds and they can be accessed by iterating from [0, activeCount()-1] (which are the indices of the active list).

It's a common practice to expose storage arrays through a `count()` method and a `read(i)` method ..

nms-7: Agree it's a little awkward. Do we have to expose this it all? `matureActive()` does not require an input, and emits the `BondMature(bond)` to log which bonds where matured anyway.

brandoniles: Ah, in the full context I see the pattern of simulating an array. (The code review interface folds unchanged code above it out of view :-/ )
```
issuedCount()
issuedBondAt(i)
activeCount()
activeBondat(i)
```
This makes more sense, I'm ok with keeping it.

brandoniles: Minor: The @notice lines should be @param lines in the natspec

brandoniles (commented):
I think this'll be the last pass of minors, thanks for bearing with it 👍

spot-contracts/contracts/_interfaces/IBondIssuer.sol
aalavandhan: I think its a fair point if we want to expose it at all (since `matureActive` doesnt require an input). 

Leaning toward removing `activeCount` and `activeBondAt` from the interface definition but keeping it in the contract.

brandoniles (approved):
LGTM

PR #141, opened 2023-04-10
Vault interface

aalavandhan:
* Standard interface definition for vaults
* Updated the deposit "function" signature (now doesn't accept the input asset)


brandoniles (approved):
LGTM

PR #150, opened 2023-05-02
Min deployment amount enforcement

aalavandhan:
* Enforcing a minimum amount to be deployed
* Guards against potential grief from dust tokens which could trigger `recoverAndRedeploy`
* Contract owner can control `minDeploymentAmt` param

### DIFF
--- a/spot-contracts/contracts/vaults/RolloverVault.sol
+++ b/spot-contracts/contracts/vaults/RolloverVault.sol
@@ -78,7 +78,7 @@ contract RolloverVault is
     //-------------------------------------------------------------------------
     // Storage
 
-    /// @notice The enforced minimum amount of assets to that have to be deployed in each iteration.
+    /// @notice Minimum amount of assets that must be deployed for a deploy operation to succeed, denominated in perps.
     /// @dev The deployment transaction reverts, if the vaults does not have sufficient usable balance
     ///      to cover the the minimum deployment amount.
     uint256 public minDeploymentAmt;
@@ -145,8 +145,8 @@ contract RolloverVault is
     }
 
     /// @notice Updates the minimum deployment amount.
-    /// @param minDeploymentAmt_ The new minimum deployment amount.
-    function updateMinDeploymentAmt(uint256 minDeploymentAmt_) public onlyOwner {
+    /// @param minDeploymentAmt_ The new minimum deployment amount, denominated in perps.
+    function updateMinDeploymentAmt(uint256 minDeploymentAmt_) external onlyOwner {
         minDeploymentAmt = minDeploymentAmt_;
     }
 
--- a/spot-contracts/contracts/vaults/RolloverVault.sol
+++ b/spot-contracts/contracts/vaults/RolloverVault.sol
@@ -12,7 +12,7 @@ import { SafeERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/
 
 import { TrancheData, TrancheHelpers, BondHelpers } from "../_utils/BondHelpers.sol";
 import { IERC20Upgradeable, IPerpetualTranche, IBondIssuer, IBondController, ITranche } from "../_interfaces/IPerpetualTranche.sol";
-import { IVault, UnexpectedAsset, UnauthorizedTransferOut, NoDeployment, DeployedCountOverLimit } from "../_interfaces/IVault.sol";
+import { IVault, UnexpectedAsset, UnauthorizedTransferOut, InsufficientDeployment, DeployedCountOverLimit } from "../_interfaces/IVault.sol";
 
 /// @notice Storage array access out of bounds.
 error OutOfBounds();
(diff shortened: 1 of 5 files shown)


spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -75,6 +75,14 @@ contract RolloverVault is
     /// @dev The maximum number of deployed assets that can be held in this vault at any given time.
     uint256 public constant MAX_DEPLOYED_COUNT = 47;
 
+    //-------------------------------------------------------------------------
+    // Storage
+
+    /// @notice The enforced minimum amount of assets to that have to be deployed in each iteration.
```
brandoniles, line 81: Something like the below? I think it's worthwhile saying how the value is denominated.
```suggestion
    /// @notice Minimum amount of assets that must be deployed for a deploy operation to succeed, denominated in perps.
```

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -136,6 +144,12 @@ contract RolloverVault is
         _unpause();
     }
 
+    /// @notice Updates the minimum deployment amount.
+    /// @param minDeploymentAmt_ The new minimum deployment amount.
```
brandoniles, line 148:
```suggestion
    /// @param minDeploymentAmt_ The new minimum deployment amount, denominated in perps.
```

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -136,6 +144,12 @@ contract RolloverVault is
...
+    function updateMinDeploymentAmt(uint256 minDeploymentAmt_) public onlyOwner {
```
brandoniles, line 149: public -> external?

brandoniles (commented):
Nice that we can reuse that same existing logic 👍 

Might seem a bit weird that it's a minimum amount of underlying asset, but denominated in perps?

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan, line 149:
```suggestion
    function updateMinDeploymentAmt(uint256 minDeploymentAmt_) external onlyOwner {
```

aalavandhan:
> Nice that we can reuse that same existing logic 👍
> 
> Might seem a bit weird that it's a minimum amount of underlying asset, but denominated in perps?

Maybe we can clarify it somewhere, but the unit of account for perps is infact the underlying asset (ie AMPL)

aalavandhan:
> Nice that we can reuse that same existing logic 👍
> 
> Might seem a bit weird that it's a minimum amount of underlying asset, but denominated in perps?

Good catch. Updating to check the underlying amounts tranched ..

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -75,6 +75,14 @@ contract RolloverVault is
     /// @dev The maximum number of deployed assets that can be held in this vault at any given time.
     uint256 public constant MAX_DEPLOYED_COUNT = 47;
 
+    //-------------------------------------------------------------------------
+    // Storage
+
+    /// @notice Minimum amount of underlying assets that must be deployed, for a deploy operation to succeed.
+    /// @dev The deployment transaction reverts, if the vaults does not have sufficient underlying tokens
+    ///      to cover the the minimum deployment amount.
```
brandoniles, line 83:
```suggestion
    ///      to cover the minimum deployment amount.
```

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -163,11 +177,12 @@ contract RolloverVault is
...
     /// @dev Its safer to call `recover` before `deploy` so the full available balance can be deployed.
-    ///      Reverts if there are no funds to deploy.
+    ///      Reverts if no funds are rolled over or if the minimum deployment threshold is not reached.
     function deploy() public override nonReentrant whenNotPaused {
-        TrancheData memory td = _tranche(perp.getDepositBond());
-        if (_rollover(perp, td) <= 0) {
-            revert NoDeployment();
+        (uint256 deployedAmt, TrancheData memory td) = _tranche(perp.getDepositBond());
+        uint256 perpsRolledOver = _rollover(perp, td);
+        if (deployedAmt <= minDeploymentAmt || perpsRolledOver <= 0) {
```
brandoniles, line 184: Do we still need to check `perpsRolledOver <= 0`?

I think this would only be the case if we roll over a small enough dust amount of underlying such that perps rounds to 0. Would we still want to revert in that case? I think we'd want it to go through.

aalavandhan: Not having this check, opens up a grief attack when external callers can just tranche the underlying and the batch job has to pay for the gas to untranche (and or retranche into a new bond).

brandoniles: Ah, nice catch. Basically enforces that we only tranche the underlying if it can immediately be used for rotations.

Might be worth a code comment

brandoniles (approved):
LGTM

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan, line 184:
```suggestion
        // NOTE: The following enforces that we only tranche the underlying if it can immediately be used for rotations.
        if (deployedAmt <= minDeploymentAmt || perpsRolledOver <= 0) {
```

PR #152, opened 2023-05-04
Added external method to return vault asset value

aalavandhan:


spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -268,6 +268,31 @@ contract RolloverVault is
         return totalAssets;
     }
 
+    /// @inheritdoc IVault
+    /// @dev The asset value is denominated in the underlying asset.
+    function getVaultAssetValue(IERC20Upgradeable token) external override returns (uint256) {
```
brandoniles, line 273: Can this be a view function?

aalavandhan: no because getAvgPrice on perp is not

brandoniles (approved):
LGTM

PR #158, opened 2023-05-17
Vault's "meld" feature

aalavandhan:
Allows users to meld their tranches with the vault's to redeem the underlying collateral

### DIFF
--- a/spot-contracts/contracts/_utils/BondTranchesHelpers.sol
+++ b/spot-contracts/contracts/_utils/BondTranchesHelpers.sol
@@ -61,7 +61,7 @@ library BondTranchesHelpers {
         uint256 minBalanceToTrancheRatio = type(uint256).max;
         uint8 i;
         for (i = 0; i < bt.tranches.length; i++) {
-            // NOTE: We round the avaiable balance down to the nearest multiple of the
+            // NOTE: We round the available balance down to the nearest multiple of the
             //       tranche ratio. This ensures that `minBalanceToTrancheRatio`
             //       can be represented without loss as a fixedPt number.
             uint256 bal = bt.tranches[i].balanceOf(u);
--- a/spot-contracts/contracts/_utils/BondTranchesHelpers.sol
+++ b/spot-contracts/contracts/_utils/BondTranchesHelpers.sol
@@ -1,0 +1,134 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.19;
+
+import { IERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
+import { ITranche } from "../_interfaces/buttonwood/ITranche.sol";
+import { MathUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/math/MathUpgradeable.sol";
+
+/// @notice Expected tranche to be part of bond.
+/// @param tranche Address of the tranche token.
+error UnacceptableTranche(ITranche tranche);
+
+struct BondTranches {
+    ITranche[] tranches;
+    uint256[] trancheRatios;
+}
+
+/**
+ *  @title BondTranchesHelpers
+ *
+ *  @notice Library with helper functions for the bond's retrieved tranche data.
+ *
+ */
+library BondTranchesHelpers {
+    using MathUpgradeable for uint256;
+
+    // Replicating value used here:
+    // https://github.com/buttonwood-protocol/tranche/blob/main/contracts/BondController.sol
+    uint256 private constant TRANCHE_RATIO_GRANULARITY = 1000;
+
+    /// @notice Iterates through the tranche data to find the seniority index of the given tranche.
+    /// @param bt The tranche data object.
+    /// @param t The address of the tranche to check.
+    /// @return the index of the tranche in the tranches array.
+    function indexOf(BondTranches memory bt, ITranche t) internal pure returns (uint8) {
+        for (uint8 i = 0; i < bt.tranches.length; i++) {
+            if (bt.tranches[i] == t) {
+                return i;
+            }
+        }
+        revert UnacceptableTranche(t);
+    }
+
+    /// @notice For a given bond's tranche data and user address, computes the maximum number of each of the bond's tranches
+    ///         the user is able to redeem before the bond's maturity. These tranche amounts necessarily match the bond's tranche ratios.
+    /// @param bt The bond's tranche data.
+    /// @param u The address to check balance for.
+    /// @return An array of tranche token balances.
+    function computeRedeemableTrancheAmounts(BondTranches memory bt, address u)
+        internal
+        view
+        returns (uint256[] memory)
+    {
+        uint256[] memory redeemableAmts = new uint256[](bt.tranches.length);
+
+        // We Calculate how many underlying assets could be redeemed from each tranche balance,
+        // assuming other tranches are not an issue, and record the smallest amount.
+        //
+        // Usually one tranche balance is the limiting factor, we first loop through to identify
+        // it by figuring out the one which has the least `trancheBalance/trancheRatio`.
+        //
+        uint256 minBalanceToTrancheRatio = type(uint256).max;
+        uint8 i;
+        for (i = 0; i < bt.tranches.length; i++) {
+            // NOTE: We round the avaiable balance down to the nearest multiple of the
+            //       tranche ratio. This ensures that `minBalanceToTrancheRatio`
+            //       can be represented without loss as a fixedPt number.
+            uint256 bal = bt.tranches[i].balanceOf(u);
+            bal = bal - (bal % bt.trancheRatios[i]);
+
+            uint256 d = bal.mulDiv(TRANCHE_RATIO_GRANULARITY, bt.trancheRatios[i]);
+            if (d < minBalanceToTrancheRatio) {
+                minBalanceToTrancheRatio = d;
+            }
+
+            // if one of the balances is zero, we return
+            if (minBalanceToTrancheRatio == 0) {
+                return (redeemableAmts);
+            }
+        }
+
+        // Now that we have `minBalanceToTrancheRatio`, we compute the redeemable amounts.
+        for (i = 0; i < bt.tranches.length; i++) {
+            redeemableAmts[i] = bt.trancheRatios[i].mulDiv(minBalanceToTrancheRatio, TRANCHE_RATIO_GRANULARITY);
+        }
+
+        return redeemableAmts;
+    }
+
+    /// @notice For a given bond's tranche data and tranche balances available, computes the maximum number of each of the bond's tranches
+    ///         the user is able to redeem before the bond's maturity.
+    ///         The returned tranche amounts necessarily match the bond's tranche ratios.
+    /// @param bt The bond's tranche data.
+    /// @param trancheBalsAvailable The tranche balance of each bond tranche available to be used for redemption.
+    /// @return An array of tranche token balances.
+    function computeRedeemableTrancheAmounts(BondTranches memory bt, uint256[] memory trancheBalsAvailable)
+        internal
+        pure
+        returns (uint256[] memory)
+    {
+        uint256[] memory redeemableAmts = new uint256[](bt.tranches.length);
+
+        // We Calculate how many underlying assets could be redeemed from each available tranche balance,
+        // assuming other tranches are not an issue, and record the smallest amount.
+        //
+        // Usually one tranche balance is the limiting factor, we first loop through to identify
+        // it by figuring out the one which has the least `trancheBalance/trancheRatio`.
+        //
+        uint256 minBalanceToTrancheRatio = type(uint256).max;
+        uint8 i;
+        for (i = 0; i < bt.tranches.length; i++) {
+            // NOTE: We round the avaiable balance down to the nearest multiple of the
+            //       tranche ratio. This ensures that `minBalanceToTrancheRatio`
+            //       can be represented without loss as a fixedPt number.
+            uint256 bal = trancheBalsAvailable[i] - (trancheBalsAvailable[i] % bt.trancheRatios[i]);
+
+            uint256 d = bal.mulDiv(TRANCHE_RATIO_GRANULARITY, bt.trancheRatios[i]);
+            if (d < minBalanceToTrancheRatio) {
+                minBalanceToTrancheRatio = d;
+            }
+
+            // if one of the balances is zero, we return
+            if (minBalanceToTrancheRatio == 0) {
+                return (redeemableAmts);
+            }
+        }
+
+        // Now that we have `minBalanceToTrancheRatio`, we compute the redeemable amounts.
+        for (i = 0; i < bt.tranches.length; i++) {
+            redeemableAmts[i] = bt.trancheRatios[i].mulDiv(minBalanceToTrancheRatio, TRANCHE_RATIO_GRANULARITY);
+        }
+
+        return redeemableAmts;
+    }
+}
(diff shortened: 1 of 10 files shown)


spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -315,6 +338,99 @@ contract RolloverVault is
...
+        // 1) We check that the bond's underlying collateral token matches the vault's underlying.
+        // 2) We validate that the bond has at least 2 tranches to perform the meld operation.
+        // 3) We validate that the bond hasn't reached maturity as matching only works through immature redemption.
+        // 4) We check if the given bond is NOT malicious, by examining if at least one of the bond's tranches
+        //    is already in the vault from previous deployments. And if the parent bond of the tranche already in the vault
+        //    is the given bond.
+        if (
+            bond.collateralToken() != address(underlying) ||
+            bt.tranches.length <= 1 ||
+            bond.isMature() ||
```
brandoniles, line 369: I think it's possible for the bond to be past the date of maturity, but still `isMature() == false` because the mature function hasn't been called yet.

I think I'm ok with this for gas savings, it just means that we're still doing a "immature redeem" on a bond that theoretically could be "mature redeemed".

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -315,6 +338,99 @@ contract RolloverVault is
...
+        // sync vault's tranche balances
+        for (uint8 i = 0; i < bt.tranches.length; i++) {
+            _syncAndRemoveDeployedAsset(bt.tranches[i]);
+        }
+
+        // sync underlying
+        _syncAsset(underlying);
+
+        // assert that the vault's TVL does not decrease after this operation
+        if (getTVL() < tvlBefore) {
```
brandoniles, line 427: Rounding errors, if any sloppy math exists in the codebase, could trip this alarm as a false positive. But I do like keeping this strict check as much as we're able, since it enforces that MUST round down on anything that leaves the vault.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -315,6 +338,99 @@ contract RolloverVault is
...
+        // 2) We validate that the bond has at least 2 tranches to perform the meld operation.
+        // 3) We validate that the bond hasn't reached maturity as matching only works through immature redemption.
+        // 4) We check if the given bond is NOT malicious, by examining if at least one of the bond's tranches
+        //    is already in the vault from previous deployments. And if the parent bond of the tranche already in the vault
+        //    is the given bond.
+        if (
+            bond.collateralToken() != address(underlying) ||
+            bt.tranches.length <= 1 ||
+            bond.isMature() ||
+            !_isDeployedBond(bond, bt)
```
brandoniles, line 370: If this check passes... do we need the first two?

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -315,6 +338,99 @@ contract RolloverVault is
...
+            // transfer tranche balance out of the user's wallet to the vault
+            trancheAmtsIn[i] = MathUpgradeable.min(trancheAmtsIn[i], trancheAmtsUsed[i]);
+            IERC20Upgradeable(bt.tranches[i]).safeTransferFrom(_msgSender(), address(this), trancheAmtsIn[i]);
+
+            // pre-calculate the user's share
+            underlyingAmt += trancheUnderlyingBalances[i].mulDiv(trancheAmtsIn[i], trancheSupplies[i]);
+        }
+
+        // ensure that the user is sending in tranches with some redeemable value.
+        if (underlyingAmt <= 0) {
```
brandoniles, line 405: This is an interesting case...

Say the user is sending in Z's that are wiped out. It technically /should/ work... It's just that the user would get nothing back, and the Vault would be holding onto the underlying from its As share.

It would be weird for someone to do this, unless they were being charitable, and especially since we expect people to be sending in As. But sending in a Z worth 0 is not that much different from a Z that's worth some positive dust amount.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -74,6 +92,10 @@ contract RolloverVault is
     /// @dev The maximum number of deployed assets that can be held in this vault at any given time.
     uint256 public constant MAX_DEPLOYED_COUNT = 47;
 
+    /// @dev The maximum percentage fee paid in the underlying assets by users who "meld".
+    ///      The meld fee paid by the user in the underlying asset.
+    uint256 public constant MAX_MELD_FEE_PERC = 5 * UNIT_PERC; // (5%)
```
brandoniles, line 97: The right value, and whether it should be configurable through governance, is worthy of some discussion.

I'm feeling like this is low, since practically speaking, in the final week with the current config and the linear dropoff, the applied fees will be very small.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -594,4 +710,30 @@ contract RolloverVault is
...
+            maturityTimestampSec > block.timestamp ? maturityTimestampSec - block.timestamp : 0
+        );
+        uint256 appliedFeePerc = MAX_MELD_FEE_PERC.mulDiv(secondsToMaturity, bondDurationSec);
+        return underlyingAmt.mulDiv(appliedFeePerc, HUNDRED_PERC);
+    }
+
+    /// @dev Checks if at least one of the bond's tranches is part of the deployed set.
+    ///      Based on this condition we can infer that the bond is NOT malicious,
+    ///      and accept other sibling tranches from this bond.
+    function _isDeployedBond(IBondController bond, BondTranches memory bt) private view returns (bool) {
```
brandoniles, line 729: Wondering if this deserves to be its own function.

It's assumed, but not checked in this context, that bt corresponds to bond. The function above this knows it, since it makes the call itself.

We could add an error check into each loop iteration, to make this function more "component like", but thinking since it's only called in one place, we can just fold this into the caller function and not worry about it.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -594,4 +710,30 @@ contract RolloverVault is
...
+        for (uint8 i = 0; i < bt.tranches.length; i++) {
+            // NOTE: We have to double check if the tranche in the deployed set
+            //       links back to the incoming bond, to ensure that the bond is not malicious.
+            if (_deployed.contains(address(bt.tranches[i])) && bt.tranches[i].bond() == address(bond)) {
```
brandoniles, line 733: Could there be a malicious bond that shares one tranche with a real bond, but also has malicious tranches? Wouldn't we need to check every tranche?

brandoniles (commented):
Initial comments on RolloverVault.sol

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: I'm leaning toward making the vault completely immutable and thus having no governance leavers at all. 

Also, there's no correct answer for what the "right" fee is. We generally feel something in the order of magnitude  of 5% for the duration makes sense. With the linear drop-off, it should be less than 1.25% in the last week. Seems like a reasonable trade-off for market actors to either hold to maturity or redeem early and pay the fee.

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: Right the 3rd check should implicitly cover the first two as well. Since we trust spot to 1) add tranches from only AMPL bonds and 2) be configured with the correct bond issuer.

We can comment 1&2 out as an optimization.

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: Yes, Its possible for the bond to be past maturity but `isMature` to be false. In which case the `bond.redeem` (ie the immature redemption) will execute as intended and not revert. The current flow should work as is. 

Alternatively we could even get rid of this check and rely on the `bond.redeem` to revert if the bond `isMature` ..

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: This also prevents against the user sending in no tranches as all.

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: With the withheld fees the vault tvl in almost all cases should be strictly higher (even with the rounding down)

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: What we really want to check here is if any of the tranches which are already in the vault, point back to the incoming bond. We assume that the tranches in the system have been vetted and their references are truthful. 

A simple but gas inefficient will look like this:
```
bool whitelistedBond = false;
for(uint8 i=0; i<_deployed.length(); i++){
  whitelistedBond = whitelistedBond ||  ITranche(_deployed.at(i)).bond() == bond;
}
return whitelistedBond;
```
 
Alternatively, we can iterate through the incoming bond's tranches to check if one of them are already in the vault. This should be more gas efficient as the tranche_len will be 2/3 and the vault_len will be much higher (and grows as a multiple of the tranche_len).

This still doesn't guard against malicious bonds. As a malicious bond fake the references to just have one "real" tranche (which is in the system). The 2nd check guards against this, the tranche in the system should link back to the actual whitelisted parent (from the spot issuer). We check if the incoming bond, matches the parent bond of the matched tranche in the system.

spot-contracts/contracts/vaults/RolloverVault.sol
brandoniles: True. Also, the vault wouldn't receive any fee for it. The risk might not be worth it for the small amount of short term liquidity that it might possibly (though very rarely, if ever) receive.

spot-contracts/contracts/vaults/RolloverVault.sol
brandoniles: I'm a little nervous that we don't really know what "reasonable" is yet. I think I'm still leaning towards governance-configurable fee in the beginning, and we can always lock it down later.

brandoniles (commented):
Looks really close 👍

spot-contracts/contracts/_utils/BondTranchesHelpers.sol
```diff
@@ -0,0 +1,134 @@
...
+        return redeemableAmts;
+    }
+
+    /// @notice For a given bond's tranche data and tranche balances available, computes the maximum number of each of the bond's tranches
+    ///         the user is able to redeem before the bond's maturity.
+    ///         The returned tranche amounts necessarily match the bond's tranche ratios.
+    /// @param bt The bond's tranche data.
+    /// @param trancheBalsAvailable The tranche balance of each bond tranche available to be used for redemption.
+    /// @return An array of tranche token balances.
+    function computeRedeemableTrancheAmounts(BondTranches memory bt, uint256[] memory trancheBalsAvailable)
```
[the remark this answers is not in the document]
aalavandhan: I like having both styles exposed by the library, internally maybe we could just use one single implementation rather than duplicating the code.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -334,6 +359,116 @@ contract RolloverVault is
...
+        }
+
+        // track the vault's TVL before the match operation
+        uint256 tvlBefore = getTVL();
+
+        // Fist we check if the vault has all the tranches for proportional redemption and if so recover the underlying.
+        // On successful redemption, The vault's individual tranche balances will decrease, and the underlying balances will increase.
+        // If the vault tranches are not eligible for proportional redemption, its a no-op.
+        // NOTE: We sync balances finally at the end of this function.
+        _execImmatureTrancheRedemption(bond, bt);
```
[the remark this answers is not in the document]
aalavandhan: Right

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -334,6 +359,116 @@ contract RolloverVault is
...
+        whenNotPaused
+        returns (uint256)
+    {
+        // get bond tranches
+        BondTranches memory bt = bond.getTranches();
+
+        // * We check that the bond's underlying collateral token matches the vault's underlying.
+        // * We validate that the bond has at least 2 tranches to perform the meld operation.
+        // * We validate that the bond hasn't reached maturity as matching only works through immature redemption.
+        if (bond.collateralToken() != address(underlying) || bt.tranches.length <= 1 || bond.isMature()) {
```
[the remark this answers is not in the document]
aalavandhan: External contracts which interact with bw tranche contracts use `isMature` to check for if the bond is mature rather than checking against the maturity date (we treat it as a black box and just interact with it using the public interface). So I'm leaning toward keeping it as is ..

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: Alternatively, the fee could be charged as a % of the incoming asset

spot-contracts/contracts/vaults/RolloverVault.sol
nms-7: Another way to do this is to reverse traverse Spot's Issuer's activeBondIDXs or issuedAt
- the bond will be active if not matured, but the bondIssuer's `issued` list is ordered by issued date, so theoretically won't have to check more than 4 indexes
- the bond will have to have been issued by spot's bond issuer

The only issue is that the BondIssuer can be updated.. but maybe we can add an 'old' issuers list? this would only ever be needed for the few bonds issued right before an update

spot-contracts/contracts/_utils/BondTranchesHelpers.sol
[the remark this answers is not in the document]
brandoniles: I see both functions are marked `internal`. Should one or both of these be public/external?

I do like the idea of not duplicating code, but it does cost more gas for one of them... No big deal if the more expensive one is not in the typical vault operations path.

spot-contracts/contracts/vaults/RolloverVault.sol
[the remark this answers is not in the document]
brandoniles: Ok, I can see that for consistency 👍

spot-contracts/contracts/vaults/RolloverVault.sol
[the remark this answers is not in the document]
brandoniles: Got it. This prioritizes AMPL holdings w/in the vault over income from melding, in some cases where the early-redeem redeems tranches that would have otherwise been melded. I don't think this would be too frequent.

The melder will pay a bit more gas than they would otherwise, to support system upkeep.

I think both of these are ok.

spot-contracts/contracts/_utils/BondTranchesHelpers.sol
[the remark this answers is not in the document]
aalavandhan: They are library functions and have to be `private`/`internal` ?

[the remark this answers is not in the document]
aalavandhan: Lets keep this version and do a separate pass to optimize for gas

spot-contracts/contracts/vaults/RolloverVault.sol
aalavandhan: Right, we don't really keep track of the list of bonds in perp anywhere. 

I think we should do an efficiency pass down the road. We can clean up the code a lot if we assume that a Bond has only 2 tranches. (and the list of tranches in perp will correspond to the set of bonds in the system)

brandoniles (approved):
LGTM

PR #159, opened 2023-05-17
Added perp redemption to vault recovery flow

aalavandhan:
- As part of the batch recovery, we first redeem the earned perp tokens for the underlying tranches and then continue with the existing flow.
- Updated perp methods "deposit", "redeem" and "rollover" to return data about the operation.

### DIFF
--- a/spot-contracts/contracts/_interfaces/IPerpetualTranche.sol
+++ b/spot-contracts/contracts/_interfaces/IPerpetualTranche.sol
@@ -38,15 +38,18 @@ interface IPerpetualTranche is IERC20Upgradeable {
     /// @notice Deposits tranche tokens into the system and mint perp tokens.
     /// @param trancheIn The address of the tranche token to be deposited.
     /// @param trancheInAmt The amount of tranche tokens deposited.
+    /// @return The amount of perp tokens minted.
     function deposit(ITranche trancheIn, uint256 trancheInAmt) external returns (uint256);
 
     /// @notice Burn perp tokens and redeem the share of reserve assets.
     /// @param perpAmtBurnt The amount of perp tokens burnt from the caller.
+    /// @return tokensOut The list of reserve tokens redeemed.
+    /// @return tokenOutAmts The list of reserve token amounts redeemed.
     function redeem(uint256 perpAmtBurnt)
         external
         returns (IERC20Upgradeable[] memory tokensOut, uint256[] memory tokenOutAmts);
 
-    struct RolloverPreview {
+    struct RolloverData {
         /// @notice The perp denominated value of tokens rolled over.
         uint256 perpRolloverAmt;
         /// @notice The amount of tokens rolled out.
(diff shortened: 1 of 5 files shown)


spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -185,6 +186,25 @@ contract RolloverVault is
 
     /// @inheritdoc IVault
     function recover() public override nonReentrant whenNotPaused {
+        // Redeem perp for tranches
+        uint256 perpBalance = perp.balanceOf(address(this));
+        if (perpBalance > 0) {
+            (IERC20Upgradeable[] memory tranchesRedeemed, ) = perp.computeRedemptionAmts(perpBalance);
```
brandoniles, line 192: Looking through the lens of efficiency...

- We don't really care about computing the redemption amounts. We really only care about getting the list of reserve assets.
- Both computeRedemptionAmts and redeem call updateState() on perp, even though we only need to do it once.

I wonder, would it be better if redeem() just returned the assets itself? Seems like it would be useful for clients, as this one shows. Then we could make just the one call.

We also may end up syncing a particular tranche more than once, if there are any in common between the perp redeem and the deployed set. The multiple syncs would probably require building up a list of assets to sync then doing them at the end, but I'm not sure the gas savings would pencil out

aalavandhan: Agreed we can just make all 3 functions, ie) deposit, redeem and rollover return the full data.

spot-contracts/contracts/_interfaces/IPerpetualTranche.sol
```diff
@@ -38,11 +38,29 @@ interface IPerpetualTranche is IERC20Upgradeable {
     /// @notice Deposits tranche tokens into the system and mint perp tokens.
     /// @param trancheIn The address of the tranche token to be deposited.
     /// @param trancheInAmt The amount of tranche tokens deposited.
-    function deposit(ITranche trancheIn, uint256 trancheInAmt) external;
+    function deposit(ITranche trancheIn, uint256 trancheInAmt) external returns (uint256);
```
brandoniles, line 41: Add a @return NatSpec?

Other places as well

spot-contracts/contracts/_interfaces/IPerpetualTranche.sol
```diff
@@ -38,11 +38,29 @@ interface IPerpetualTranche is IERC20Upgradeable {
...
 
     /// @notice Burn perp tokens and redeem the share of reserve assets.
     /// @param perpAmtBurnt The amount of perp tokens burnt from the caller.
-    function redeem(uint256 perpAmtBurnt) external;
+    function redeem(uint256 perpAmtBurnt)
+        external
+        returns (IERC20Upgradeable[] memory tokensOut, uint256[] memory tokenOutAmts);
+
+    struct RolloverPreview {
```
brandoniles, line 49: Minor, but wonder if this should be renamed since it's also a rollover return data payload.
If we do, maybe save it for just a rename PR.

aalavandhan: We can rename it to `RolloverData`

nms-7 (approved):

brandoniles (approved):
LGTM

PR #170, opened 2023-08-09
Fixed rounding issue with recovery

aalavandhan:
The `computeRedeemableTrancheAmounts` should round the balances down to a number divisible by the tranche ratio. 

For example, if the tranche balances are "10, 15, 7.461048491123254231" the previous implementation returned "2.984419396449301692, 4.476629094673952538, 7.461048491123254231".

However, this is not the right redemption ratio, which is expected to be precisely divisible. 
https://github.com/buttonwood-protocol/tranche/blob/main/contracts/BondController.sol#L229

The proposed change rounds down to achieve amounts are consistent with the expected tranche ratio. 
We get -> "2.984419396449301600, 4.476629094673952400, 7.461048491123254000"

### DIFF
--- a/spot-contracts/contracts/_utils/BondHelpers.sol
+++ b/spot-contracts/contracts/_utils/BondHelpers.sol
@@ -210,24 +210,35 @@ library BondHelpers {
         BondTranches memory bt = getTranches(b);
         uint256[] memory redeemableAmts = new uint256[](bt.tranches.length);
 
-        // Calculate how many underlying assets could be redeemed from each tranche balance,
+        // We Calculate how many underlying assets could be redeemed from each tranche balance,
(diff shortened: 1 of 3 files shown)


aalavandhan:
Consider another example. This time lets just consider fixed point numbers.

For tranche ratios of 200,300,1000 and granularity=1000.

Lets assume balances are "[1000, 5001, 503]"

As per the previous implementation, the redeemable tranche ratios returned will be "[]"

1000, 5001, 503

201


For example, if the tranche balances are "10, 15, 7.461048491123254231" the previous implementation returned "2.984419396449301692, 4.476629094673952538, 7.461048491123254231".

brandoniles (approved):
Nice solution. I agree this strikes the right balance for gas efficiency.

PR #173, opened 2023-08-21
hot-fix vault rename

aalavandhan:
* Imported all oz dependencies into the project
* Updated the underlying ERC-20 library making name, symbol internal variables
* Added setter to update the name and symbol

### DIFF
--- a/spot-contracts/.openzeppelin/goerli.json
+++ b/spot-contracts/.openzeppelin/goerli.json
@@ -3442,6 +3442,267 @@
           }
         }
       }
+    },
+    "434e3f9e5ca89604923f560238c1549d17341d5000a67900fb3b705d5348ddd7": {
+      "address": "0x795F55f37bd9ab8fa361D0FC0eAd6ff036aebA44",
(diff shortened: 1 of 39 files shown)


brandoniles (commented):
IIUC, the old name slot has reserved at least 26 chars, and the new one is 17
The old symbol has reserved at least 12 chars, the new one is 6
So no worries about accidental overruns.

There might be some remnants of the old strings after the new ones, but since these are smaller everything should be ok.

I noticed you flattened the directory structure of the OZ libs, any reason why? I assume you modified the import paths in the OZ files to accomodate. I guess you would have had to modify somehow anyway unless the `@openzeppelin` prefix was changed. Not sure how that works in the tooling though.

In any case, looks good to me as it is.

Are the tests passing for you locally? Github suite is failing at the lint stage

aalavandhan:
> I noticed you flattened the directory structure of the OZ libs, any reason why? I assume you modified the import paths in the OZ files to accomodate. I guess you would have had to modify somehow anyway unless the `@openzeppelin` prefix was changed. Not sure how that works in the tooling though.

I think having both dependencies (oz external and internal) seems to be causing variable name clashes.

brandoniles (approved):
LGTM

PR #177, opened 2023-09-13
Fee strategy 2.0

aalavandhan:
Moving for fixed-percentage based fees one which is computed on the fly based on `TVL` in the SPOT and supporting rollover vault.


spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -909,22 +862,25 @@ contract PerpetualTranche is
...
+        uint256 feePerc = feeStrategy.computeMintFeePerc();
         uint256 totalSupply_ = totalSupply();
         uint256 stdTrancheInAmt = _toStdTrancheAmt(trancheInAmt, computeDiscount(trancheIn));
         uint256 trancheInPrice = computePrice(trancheIn);
         uint256 perpAmtMint = (totalSupply_ > 0)
             ? (stdTrancheInAmt * trancheInPrice).mulDiv(totalSupply_, _reserveValue())
             : stdTrancheInAmt.mulDiv(trancheInPrice, UNIT_PRICE);
-        return (perpAmtMint);
+        // NOTE: The mint fees are settled by simply minting lesser perps.
+        perpAmtMint -= perpAmtMint.mulDiv(feePerc, HUNDRED_PERC);
```
aalavandhan, line 874: Round up here?

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -934,17 +890,24 @@ contract PerpetualTranche is
             redemptionAmts[i] = (totalSupply_ > 0)
                 ? _reserveBalance(reserveTokens[i]).mulDiv(perpAmtBurnt, totalSupply_)
                 : 0;
+            // NOTE: The burn fees are settled by simply redeeming lesser tranches.
+            redemptionAmts[i] -= redemptionAmts[i].mulDiv(feePerc, HUNDRED_PERC);
```
aalavandhan, line 894: Round up here as well?

brandoniles (commented):
First pass comments, but haven't looked at PerpetualTranche or FeeStrategy yet

So far looks good

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -959,13 +922,28 @@ contract PerpetualTranche is
...
         r.trancheInAmt = trancheInAmtAvailable;
         uint256 stdTrancheInAmt = _toStdTrancheAmt(trancheInAmtAvailable, trancheInDiscount);
-
-        // Basic rollover:
-        // (stdTrancheInAmt . trancheInPrice) = (stdTrancheOutAmt . trancheOutPrice)
         uint256 stdTrancheOutAmt = stdTrancheInAmt.mulDiv(trancheInPrice, trancheOutPrice);
-        r.trancheOutAmt = _fromStdTrancheAmt(stdTrancheOutAmt, trancheOutDiscount);
+
+        // A postive fee percentage implies that perp charges rotators by
+        // accepting tranchesIn at a discount, ie) lesser tranches out.
```
brandoniles, line 936:
```suggestion
        // accepting tranchesIn at a discount, ie) fewer tranches out.
```

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -959,13 +922,28 @@ contract PerpetualTranche is
...
+        // Given the amount of tranches In, we compute the amount of tranches out
         r.trancheInAmt = trancheInAmtAvailable;
         uint256 stdTrancheInAmt = _toStdTrancheAmt(trancheInAmtAvailable, trancheInDiscount);
-
-        // Basic rollover:
-        // (stdTrancheInAmt . trancheInPrice) = (stdTrancheOutAmt . trancheOutPrice)
         uint256 stdTrancheOutAmt = stdTrancheInAmt.mulDiv(trancheInPrice, trancheOutPrice);
-        r.trancheOutAmt = _fromStdTrancheAmt(stdTrancheOutAmt, trancheOutDiscount);
+
+        // A postive fee percentage implies that perp charges rotators by
```
brandoniles, line 935:
```suggestion
        // A positive fee percentage implies that perp charges rotators by
```

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -959,13 +922,28 @@ contract PerpetualTranche is
...
+        // accepting tranchesIn at a discount, ie) lesser tranches out.
+        if (feePerc > 0) {
```
brandoniles, line 937: I guess it's worth asking if these two conditions could be collapsed into one.

`feePerc` is already a signed number, so could we have one condition that simply does `HUNDRED_PERC + feePerc`?

But I'd keep the comments, those are very helpful.

brandoniles: Maybe it just collapses into line 933

However, if we round up in one case, and down in the other, then we should keep them separate though

brandoniles (commented):
This covers PerpetualTranche.sol

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,226 @@
...
+    using BondHelpers for IBondController;
+    using PerpHelpers for IPerpetualTranche;
+
+    /// @dev The returned fee percentages are fixed point numbers with {PERC_DECIMALS} places.
+    ///      The decimals should line up with value expected by consumer (perp).
+    uint8 public constant PERC_DECIMALS = 8;
+    uint256 public constant HUNDRED_PERC = 10**PERC_DECIMALS; // 100%
+    uint256 public constant MAX_PERC = HUNDRED_PERC / 10; // 10%
+
+    /// @dev Number of seconds in one year.
```
nms-7, line 55: pedantic, but we could say 365 1/4 days per year to account for leap years

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,226 @@
...
+
+    // Replicating value used here:
+    // https://github.com/buttonwood-protocol/tranche/blob/main/contracts/BondController.sol
+    uint256 private constant TRANCHE_RATIO_GRANULARITY = 1000;
+
+    /// @notice Reference to the perpetual token.
+    IPerpetualTranche public immutable perp;
+
+    /// @notice Reference to the rollover vault.
+    IVault public immutable vault;
```
nms-7, line 66: not necessary now, but if we ever had multiple vaults we might want this to be a list

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,226 @@
...
+    ///        as a fixed point number with {PERC_DECIMALS} places.
+    function updateBurnFees(uint256 maxBurnFeePerc_) external onlyOwner {
+        require(maxBurnFeePerc_ <= HUNDRED_PERC, "FeeStrategy: burn fee too high");
+        maxBurnFeePerc = maxBurnFeePerc_;
+    }
+
+    /// @notice Update the parameters determining the slope and asymptotes of the sigmoid fee curve.
+    /// @param p Lower, Upper and Growth sigmoid paramters are fixed point numbers with {PERC_DECIMALS} places.
+    function updateRolloverFees(SigmoidParams memory p) external onlyOwner {
+        require(p.lower >= -int256(MAX_PERC), "FeeStrategy: fee bound too low");
```
nms-7, line 136: add `require(p.lower < p.upper`, ...);

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,226 @@
...
+
+    /// @notice Reference to the perpetual token.
+    IPerpetualTranche public immutable perp;
+
+    /// @notice Reference to the rollover vault.
+    IVault public immutable vault;
+
+    /// @notice Defines the neutral range inside which we ignore any deviation between
+    ///         the expected and current TVLs.
+    uint256 public deviationTresholdPerc;
```
nms-7, line 70:
```suggestion
    uint256 public deviationThresholdPerc;
```

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,226 @@
...
+        return PERC_DECIMALS;
+    }
+
+    /// @notice Computes the ratio between the vault's TVL and the target TVL.
+    /// @dev Adjusts the computed result based on the `deviationTresholdPerc`.
+    /// @return The ratio as a fixed point number with {PERC_DECIMALS} places.
+    function computeDeviationRatio(IBondController referenceBond) public returns (uint256) {
+        // NOTE: Ensure that the perp's TVL and vault's TVL have the same base denomination.
+        (uint256 perpRatio, ) = perp.computeEffectiveTrancheRatio(referenceBond);
+        uint256 targetTVL = perp.getTVL().mulDiv(TRANCHE_RATIO_GRANULARITY, perpRatio);
```
nms-7, line 182: so we're committing to needing 4x rather than 3x in the vault for 25:75 bonds?

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -1182,24 +1122,6 @@ contract PerpetualTranche is
         }
     }
 
-    /// @dev Enforces that supply strictly reduces after the redemption.
-    function _enforceSupplyReduction(uint256 prevSupply) private view {
```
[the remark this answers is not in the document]
aalavandhan: We had this to ensure than on redemption supply strictly reduces, as we could mint perps as redemption fees. However, we don't really have perp tokens fees anymore.

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -442,6 +429,9 @@ contract PerpetualTranche is
         if (address(feeStrategy_) == address(0)) {
             revert UnacceptableReference();
         }
+        if (feeStrategy_.decimals() != PERC_DECIMALS) {
+            revert InvalidStrategyDecimals(feeStrategy_.decimals(), PERC_DECIMALS);
```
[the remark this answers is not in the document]
aalavandhan: I'm thinking of just getting rid of error parameters in an upcoming PR. I don't think they add any value ..

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -629,18 +594,12 @@ contract PerpetualTranche is
 
         // verifies if rollover amount is acceptable
         if (r.trancheInAmt == 0 || r.tokenOutAmt == 0 || r.perpRolloverAmt == 0) {
-            revert UnacceptableRolloverAmt(r.trancheInAmt, r.tokenOutAmt, r.perpRolloverAmt);
+            return r;
```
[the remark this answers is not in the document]
aalavandhan: Right, the rollover vault "tries" to rollover every available tranche and its preferable for the rollover not to revert if there aren't enough tokens available. If not we'll have to do some exceptional handling in the vault.

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -872,14 +830,9 @@ contract PerpetualTranche is
...
 
-    /// @inheritdoc IPerpetualTranche
-    function protocolFeeCollector() public view override returns (address) {
-        return owner();
-    }
-
     /// @inheritdoc IPerpetualTranche
     function feeToken() public view override returns (IERC20Upgradeable) {
-        return feeStrategy.feeToken();
+        return IERC20Upgradeable(address(this));
```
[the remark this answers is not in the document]
aalavandhan: Its being used by the `RouterV1`. We can get rid of this after we update the router ..

spot-contracts/contracts/_deprecated/BasicFeeStrategy.sol
```diff
@@ -120,4 +120,24 @@ contract BasicFeeStrategy is IFeeStrategy, OwnableUpgradeable {
         uint256 absoluteFee = (rolloverFeePerc.abs() * rolloverAmt) / HUNDRED_PERC;
         return (rolloverFeePerc.sign() * absoluteFee.toInt256(), 0);
     }
+
+    /// @inheritdoc IFeeStrategy
+    function computeMintFeePerc() external pure override returns (uint256) {
```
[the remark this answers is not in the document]
aalavandhan: Will do it in a later PR

spot-contracts/contracts/_utils/Sigmoid.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// https://github.com/ampleforth/ampleforth-contracts/blob/master/contracts/UFragmentsPolicy.sol
+pragma solidity ^0.8.19;
+
+import { SafeMathInt } from "ampleforth-contracts/contracts/lib/SafeMathInt.sol";
```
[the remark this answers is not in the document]
aalavandhan: We import this mainly for `SafeMathInt.twoPower`. However we could just port that over and rely on safe math for the rest of the math. Will do it in a later PR

spot-contracts/contracts/_interfaces/IFeeStrategy.sol
```diff
@@ -2,38 +2,40 @@
...
 
-    /// @notice Computes the fees while minting given amount of perp tokens.
-    /// @dev The mint fee can be either positive or negative. When positive it's paid by the minting users to the reserve.
-    ///      When negative its paid to the minting users by the reserve.
-    ///      The protocol fee is always non-negative and is paid by the users minting to the
-    ///      perp contract's fee collector.
-    /// @param amount The amount of perp tokens to be minted.
-    /// @return reserveFee The fee paid to the reserve to mint perp tokens.
-    /// @return protocolFee The fee paid to the protocol to mint perp tokens.
+    /// @notice DEPRECATED.
```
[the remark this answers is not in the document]
aalavandhan: The router still uses the functions. Will remove it when the router is updated ..

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -228,9 +214,8 @@ contract PerpetualTranche is
     uint8 public constant PRICE_DECIMALS = 8;
     uint256 public constant UNIT_PRICE = (10**PRICE_DECIMALS);
 
-    uint8 public constant PERC_DECIMALS = 6;
-    uint256 public constant UNIT_PERC = 10**PERC_DECIMALS;
-    uint256 public constant HUNDRED_PERC = 100 * UNIT_PERC;
+    uint8 public constant PERC_DECIMALS = 8;
+    uint256 public constant HUNDRED_PERC = 10**PERC_DECIMALS;
```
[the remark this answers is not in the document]
aalavandhan: 100% when represented as decimals = 1.0

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -909,22 +862,25 @@ contract PerpetualTranche is
     // Private methods
 
     /// @dev Computes the perp mint amount for given amount of tranche tokens deposited into the reserve.
-    function _computeMintAmt(ITranche trancheIn, uint256 trancheInAmt) private view returns (uint256) {
+    function _computeMintAmt(ITranche trancheIn, uint256 trancheInAmt) private returns (uint256) {
```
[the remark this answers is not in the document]
aalavandhan: Right, we could move to a poke model and hook it up to the orchestrator

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,226 @@
...
+ *             and is a fixed percentage fee set by the owner.
+ *
+ *          2) The burn fees are turned on when deviationFactor > 1,
+ *             and is a fixed percentage fee set by the owner.
+ *
+ *          3) The rollover fees are signed and can flow in either direction.
+ *             The fee is a percentage of the tranches rolled over and is computed
+ *             through a sigmoid function. The slope and asymptotes are set by the owner.
+ *
+ *              rotationsPerYear = mintingBondDuration / 1_year
```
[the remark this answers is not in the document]
aalavandhan: However, `maxMaturityDuration` can be updated anytime by the owner. Its safer to rely on a bond which is already in the system (ie the deposit bond).

spot-contracts/contracts/strategies/FeeStrategy.sol
aalavandhan: We could swap out the fee strategy at that point. Okay keeping it as is rn.

spot-contracts/contracts/strategies/FeeStrategy.sol
aalavandhan: For now that is the case. Once we allow for flash-rollovers, we need to revisit this

spot-contracts/contracts/strategies/FeeStrategy.sol
nms-7, line 37: should be the opposite then right?
```suggestion
 *              rotationsPerYear = 1_year / mintingBondDuration
```

nms-7, line 37: the math in the actual code below works because 1 / rotationsPerYear = mintingBondDuration / 1_year, but this should read 
```suggestion
 *              rotationsPerYear = 1_year / mintingBondDuration
```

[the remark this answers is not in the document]
nms-7: Good point. At least this is bounded between `maxMaturityDuration` and `maxMaturityDuration - issueWindowOffset`. Since the issuer at the perp level, and the issuer params at the issuer level are both configurable though, in practice we've just updated the issuer – we may want to consider making the issuer params  immutable in a future update

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+        require(p.upper <= int256(MAX_PERC), "FeeStrategy: fee bound too high");
+        require(p.lower < p.upper, "FeeStrategy: paramters invalid");
+        rolloverFeeAPR.lower = p.lower;
+        rolloverFeeAPR.upper = p.upper;
+        rolloverFeeAPR.growth = p.growth;
+    }
+
+    /// @inheritdoc IFeeStrategy
+    function computeMintFeePerc() external override returns (uint256) {
+        // when vault tvl > target, we encorage supply growth by dropping mint fees to 0
```
brandoniles, line 146:
```suggestion
        // when vault tvl > target, we encourage supply growth by dropping mint fees to 0
```

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+        require(p.lower >= -int256(MAX_PERC), "FeeStrategy: fee bound too low");
+        require(p.upper <= int256(MAX_PERC), "FeeStrategy: fee bound too high");
+        require(p.lower < p.upper, "FeeStrategy: paramters invalid");
+        rolloverFeeAPR.lower = p.lower;
+        rolloverFeeAPR.upper = p.upper;
+        rolloverFeeAPR.growth = p.growth;
+    }
+
+    /// @inheritdoc IFeeStrategy
+    function computeMintFeePerc() external override returns (uint256) {
```
brandoniles, line 145: Could this be a view function?

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+        return (rolloverAPR * referenceBond.duration().toInt256()) / ONE_YEAR_SEC;
+    }
+
+    /// @inheritdoc IFeeStrategy
+    function decimals() external pure override returns (uint8) {
+        return PERC_DECIMALS;
+    }
+
+    /// @notice Computes the ratio between the vault's TVL and the target TVL.
+    /// @dev Adjusts the computed result based on the `deviationThresholdPerc`.
```
brandoniles, line 178: Worth documenting: A "perfect" desired deviation ratio-- is it 1.0? or 0.0?

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+    /// @return The ratio as a fixed point number with {PERC_DECIMALS} places.
+    function computeDeviationRatio(IBondController referenceBond) public returns (uint256) {
+        // NOTE: Ensure that the perp's TVL and vault's TVL have the same base denomination.
+        (uint256 perpRatio, ) = perp.computeEffectiveTrancheRatio(referenceBond);
+        uint256 targetTVL = perp.getTVL().mulDiv(TRANCHE_RATIO_GRANULARITY, perpRatio);
+        uint256 deviationFactor = vault.getTVL().mulDiv(HUNDRED_PERC, targetTVL);
+
+        // For example, the threshold is set to 5%.
+        // When the computed ratio is below 1, we leave it unchanged.
```
brandoniles, line 187: Not sure I follow this logic.

So, 5% is not +/-5%?
I.e. If it's 0.97, should it up moved up to 1.0?

I'd expect (0.95, 1.05) -> 1.0, otherwise, leave it unchanged.

aalavandhan: I think in the context of the rotation vault, 0.97 isn't good enough as it wont rollover the entire supply.
What we really want is 1.0 or more, thus the neutral band is defined as (1.0 - 1.05) ..

spot-contracts/contracts/strategies/FeeStrategy.sol
aalavandhan: Since perp's `getTVL` calls `updateState` this can't be a view function

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+        require(maxBurnFeePerc_ <= HUNDRED_PERC, "FeeStrategy: burn fee too high");
+        maxBurnFeePerc = maxBurnFeePerc_;
+    }
+
+    /// @notice Update the parameters determining the slope and asymptotes of the sigmoid fee curve.
+    /// @param p Lower, Upper and Growth sigmoid paramters are fixed point numbers with {PERC_DECIMALS} places.
+    function updateRolloverFees(SigmoidParams calldata p) external onlyOwner {
+        require(p.lower >= -int256(MAX_PERC), "FeeStrategy: fee bound too low");
+        require(p.upper <= int256(MAX_PERC), "FeeStrategy: fee upper bound too high");
+        require(p.lower < p.upper, "FeeStrategy: paramters invalid");
```
aalavandhan, line 138:
```suggestion
        require(p.lower <= p.upper, "FeeStrategy: paramters invalid");
```

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+ *             and is a fixed percentage fee set by the owner.
+ *
+ *          2) The burn fees are turned on when deviationFactor > 1,
+ *             and is a fixed percentage fee set by the owner.
+ *
+ *          3) The rollover fees are signed and can flow in either direction.
+ *             The fee is a percentage of the tranches rolled over and is computed
+ *             through a sigmoid function. The slope and asymptotes are set by the owner.
+ *
+ *              rotationsPerYear = 1_year / mintingBondDuration
```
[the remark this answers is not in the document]
aalavandhan: Right, however that can be changed by the owner any time. It safer to look at tranches in (or about to enter) the system, ie) ones belonging to the deposit bond.

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+    ///         the expected and current TVLs.
+    uint256 public deviationThresholdPerc;
+
+    /// @notice The perp token's mint fee percentage celing.
+    uint256 public maxMintFeePerc;
+
+    /// @notice The perp token's burn fee percentage celing.
+    uint256 public maxBurnFeePerc;
+
+    struct SigmoidParams {
```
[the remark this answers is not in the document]
aalavandhan: No, Ampleforth supply policy works the same way as this contract. 
```
(new supply = old_supply * deviationFactor)
```
A deviation factor of 1.0, means the supply does not change ..

spot-contracts/contracts/strategies/FeeStrategy.sol
aalavandhan: Perfect is 1.0. Adding a comment

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+    using MathUpgradeable for uint256;
+    using SafeCastUpgradeable for uint256;
+    using BondHelpers for IBondController;
+    using PerpHelpers for IPerpetualTranche;
+
+    /// @dev The returned fee percentages are fixed point numbers with {PERC_DECIMALS} places.
+    ///      The decimals should line up with value expected by consumer (perp).
+    uint8 public constant PERC_DECIMALS = 8;
+    uint256 public constant HUNDRED_PERC = 10**PERC_DECIMALS; // 100%
+    uint256 public constant MAX_PERC = HUNDRED_PERC / 10; // 10%
```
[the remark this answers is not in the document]
aalavandhan: renamed the variable

spot-contracts/contracts/strategies/FeeStrategy.sol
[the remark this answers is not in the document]
brandoniles: Ah you're right, my mistake

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+    /// @param vault_ Reference to the vault.
+    constructor(IPerpetualTranche perp_, IVault vault_) {
+        perp = perp_;
+        vault = vault_;
+    }
+
+    /// @notice Contract initializer.
+    function init() public initializer {
+        __Ownable_init();
+        maxMintFeePerc = HUNDRED_PERC / 40; // 2.5%
```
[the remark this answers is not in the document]
brandoniles: Weird, we do it on line 53. Is there an error message?

spot-contracts/contracts/strategies/FeeStrategy.sol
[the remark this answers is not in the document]
aalavandhan: hmm. Thats right. 

The owner can update `minIssueTimeIntervalSec` anytime and it feels safer to compute this from tranches in the system.

This also adds some weirdness as `bond.duration`, is the difference between time of issue and maturity. The time of issue floats, and is locked in only when `updateState` executes and in practice could be lower than `minIssueTimeIntervalSec`.

spot-contracts/contracts/strategies/FeeStrategy.sol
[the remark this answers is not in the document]
aalavandhan: Do we? I dont think we use floating point numbers. 

For example you get something like this ..

```
 Type rational_const 3 / 2 is not implicitly convertible to expected type uint256. Try converting to type ufixed8x1 or use an explicit conversion
 
 uint256 public constant HUNDRED_PERC = 1.5 * 10**PERC_DECIMALS; // 100%, or 1.00

```

aalavandhan:
Updated the sigmoid function to have a configurable target. Previously it was hard-coded to be at 1.0, now we can just pass in the target.

We could start off with something like this. lower=-0.02, upper=0.075, growth=3, target=1.05

fee-sigmoid

brandoniles:
Shifting the line is a neat idea. Not what I expected, but I guess cleaner & more efficient than computing the ratios in front of the sigmoid.

aalavandhan:
> Shifting the line is a neat idea. Not what I expected, but I guess cleaner & more efficient than computing the ratios in front of the sigmoid.

This is what we do implicitly in the policy too. As the target rate changes, we just move sigmoid_inv(0) ..

https://github.com/ampleforth/ampleforth-contracts/blob/master/contracts/UFragmentsPolicy.sol#L329

aalavandhan:
Normalizing the input before passing to the sigmoid function or inside should in theory give us the same output. 
```
ie) x = 1.5, target = 1.05
=> sigmoid(x, target, ..params) = sigmoid(x/target, 1, ..params)
```

Given that that's the case, leaning toward not changing the sigmoid code (using the policy version as is) but just passing the `normalizedDeviation` to it. (ie currentDeviation / targetDeviation)

spot-contracts/contracts/strategies/FeeStrategy.sol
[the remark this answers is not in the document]
aalavandhan: Lets move to using `minIssueTimeIntervalSec`. Will update it in a later PR

spot-contracts/contracts/strategies/FeeStrategy.sol
```diff
@@ -0,0 +1,227 @@
...
+            computeNormalizedDeviation(referenceBond).toInt256(),
+            rolloverFeeAPR.lower,
+            rolloverFeeAPR.upper,
+            rolloverFeeAPR.growth,
+            ONE.toInt256()
+        );
+
+        // We calculate the rollover fee for the given cycle by dividing the annualized rate
+        // by the number of cycles in any given year.
+        return (rolloverAPR * referenceBond.duration().toInt256()) / ONE_YEAR_SEC;
```
aalavandhan, line 175:
```suggestion
        // TODO: use the bond duration from the bondIssuer rather than the deposit bond
        return (rolloverAPR * referenceBond.duration().toInt256()) / ONE_YEAR_SEC;
```

spot-contracts/contracts/_utils/Sigmoid.sol
```diff
@@ -20,41 +20,40 @@ library Sigmoid {
...
-    /// @return The computed value of sigmoid(x) as fixed-point number with the provided number of decimal places.
+    /// @param target The target value of "x" for which sigmoid(target) returns zero 0.
+    /// @return The computed value of sigmoid(x).
     function compute(
         int256 x,
         int256 lower,
         int256 upper,
         int256 growth,
-        uint8 decimals
+        int256 target
```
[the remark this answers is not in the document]
brandoniles: Resolving, since this is reverted back to original math

brandoniles (approved):
LGTM

PR #178, opened 2023-09-15
Vault V2; clean up, swap and deposit2

aalavandhan:
* Handing zero deposit redeem ; Added a keeper role and plausibility; Added owner controllable fees (and a few other hyper params) to the vault; Using `uint8` for iterating over deployed assets; Reorg internal functions
* Deducting a "deploymentFee" on every successful deployment
* Added redemption fees
* Added swap function
* Added `deposit2`


aalavandhan:
Broke up into #182, #183, #184, #185

PR #181, opened 2023-11-18
Cleanup

aalavandhan:
* Deleted unused contracts
* Imported `twoPow` fn into sigmoid code, removed safemath
* replaced `== 0` with `<= 0`


spot-contracts/contracts/_utils/Sigmoid.sol
```diff
@@ -2,18 +2,56 @@
...
- *  TODO: port over 2-pow and use plain math instead of the "safe math" lib.
- *
  */
 library Sigmoid {
-    using SafeMathInt for int256;
+    using SignedMathUpgradeable for int256;
+    using SafeCastUpgradeable for uint256;
+
+    /// @notice Computes 2^exp with limited precision where -100 <= exp <= 100 * one
+    /// @param one 1.0 represented in the same fixed point number format as exp
```
nms-7, line 19: swap @param lines to match ordering in function definition (`exp`, `one`)

nms-7 (approved):

spot-contracts/contracts/_utils/Sigmoid.sol
```diff
@@ -32,29 +70,25 @@ library Sigmoid {
...
+        int256 exponent = (growth * delta) / one;
+
         // Cap exponent to guarantee it is not too big for twoPower
-        if (exponent > one.mul(100)) {
-            exponent = one.mul(100);
-        }
-        if (exponent < one.mul(-100)) {
-            exponent = one.mul(-100);
-        }
+        exponent = SignedMathUpgradeable.min(SignedMathUpgradeable.max(exponent, -100 * one), 100 * one);
```
brandoniles, line 80: Is this clearer than the original version? I think I find it easier to understand the conditional form. I'm guessing it's also more gas efficient, since there are no jumps.

spot-contracts/contracts/_utils/Sigmoid.sol
```diff
@@ -2,18 +2,56 @@
...
 library Sigmoid {
-    using SafeMathInt for int256;
+    using SignedMathUpgradeable for int256;
+    using SafeCastUpgradeable for uint256;
+
+    /// @notice Computes 2^exp with limited precision where -100 <= exp <= 100 * one
+    /// @param one 1.0 represented in the same fixed point number format as exp
+    /// @param exp The power to raise 2 to -100 <= exp <= 100 * one
+    /// @return 2^exp represented with same number of decimals after the point as one
+    function twoPower(int256 exp, int256 one) internal pure returns (int256) {
```
brandoniles, line 22: If we copy this code over from ampleforth-contracts, are there any test files we can bring over with it? If so, I think we should

spot-contracts/contracts/_utils/Sigmoid.sol
aalavandhan: sure will go back

spot-contracts/contracts/_utils/Sigmoid.sol
aalavandhan: Sure .. imported

brandoniles (approved):
LGTM

PR #185, opened 2023-11-21
Added deposit2

aalavandhan:
Ability to mint spot and stampl together

### DIFF
--- a/spot-contracts/contracts/vaults/RolloverVault.sol
+++ b/spot-contracts/contracts/vaults/RolloverVault.sol
@@ -301,7 +301,7 @@ contract RolloverVault is
 
     /// @inheritdoc IVault
     /// @dev Its safer to call `recover` before `deploy` so the full available balance can be deployed.
-    ///      The vault holds `minUnderlyingBal` as underlying tokens and deploys the rest. 
+    ///      The vault holds `minUnderlyingBal` as underlying tokens and deploys the rest.
     ///      Reverts if no funds are rolled over or enforced deployment threshold is not reached.
     function deploy() public override nonReentrant whenNotPaused {
         _deductProtocolFee();
@@ -316,7 +316,7 @@ contract RolloverVault is
             revert InsufficientDeployment();
         }
 
-        // NOTE: We only trust incoming tranches from perp's deposit bond, 
+        // NOTE: We only trust incoming tranches from perp's deposit bond,
         // or ones rolled out of perp.
         IBondController depositBond = perp.getDepositBond();
         BondTranches memory bt = depositBond.getTranches();
@@ -434,7 +434,12 @@ contract RolloverVault is
         uint256 seniorsMinted = bt.tranches[0].balanceOf(address(this)) - seniorBalBefore;
         _checkAndApproveMax(bt.tranches[0], address(perp), seniorsMinted);
         uint256 perpsMinted = perp.deposit(bt.tranches[0], seniorsMinted);
+
+        // sync vault tranche balances
         _syncAndRemoveDeployedAsset(bt.tranches[0]);
+        for (uint8 i = 1; i < bt.tranches.length; i++) {
+            _syncAndAddDeployedAsset(bt.tranches[i]);
+        }
 
         // transfer perps back to user
         IERC20Upgradeable(address(perp)).safeTransfer(_msgSender(), perpsMinted);
--- a/spot-contracts/contracts/vaults/RolloverVault.sol
+++ b/spot-contracts/contracts/vaults/RolloverVault.sol
@@ -404,6 +404,53 @@ contract RolloverVault is
         return notes;
     }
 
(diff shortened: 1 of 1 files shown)


spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -404,6 +404,53 @@ contract RolloverVault is
...
+        uint256 seniorBalBefore = bt.tranches[0].balanceOf(address(this));
+
+        // transfer underlying in
+        underlying.safeTransferFrom(_msgSender(), address(this), amount);
+
+        // tranches underlying and updates vault assets with fresh tranches
+        _tranche(depositBond, bt, amount);
+
+        // mint perps using freshly tranched seniors
+        uint256 seniorsMinted = bt.tranches[0].balanceOf(address(this)) - seniorBalBefore;
```
brandoniles, line 434: It seems philosophically safer to not ever interact with the vaults existing tranche balances at all. Until the final upkeep, anyway, once the main operations are complete. This closes down a whole class of bugs & attacks.

When we tranche the underlying, shouldn't that function be able to return the number of tranches created? If that means not using the `_tranche` helper, that seems ok to me. That helper also updates the vault assets with intermediate values that we don't want anyway.

aalavandhan: In the current version, the tranche function does not return the number of tranches created. We'll have to figure that out based on the balance changes ..

https://github.com/buttonwood-protocol/tranche/blob/main/contracts/BondController.sol#L123

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -404,6 +404,53 @@ contract RolloverVault is
...
+        _checkAndApproveMax(bt.tranches[0], address(perp), seniorsMinted);
+        uint256 perpsMinted = perp.deposit(bt.tranches[0], seniorsMinted);
+        _syncAndRemoveDeployedAsset(bt.tranches[0]);
```
aalavandhan, line 437: We need to call `_syncAndAddDeployedAsset` on the other tranches

spot-contracts/contracts/vaults/RolloverVault.sol
brandoniles: Ok, in that case let's stick with your original implementation

spot-contracts/contracts/vaults/RolloverVault.sol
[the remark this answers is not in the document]
aalavandhan: We initially had a function which did either add or remove as needed. We move toward 2 explicit functions ie) add-as-needed or remove-as-needed so we can be certain to call either when deployed assets enter or leave the vault. 

As a rule of thumb In the case deployed assets leave we call `_syncAndRemoveDeployedAsset` and when they enter we call `_syncAndAddDeployedAsset`. For gas efficiency we can choose not to call either if we know that a given operation isn't going to change the vault's tranche balance.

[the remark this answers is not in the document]
aalavandhan: But yea I agree, we can `_syncAndAddDeployedAsset` all tranches after trancing the AMPL, and `_syncAndAddDeployedAsset` the senior after perp minting.

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -404,6 +404,58 @@ contract RolloverVault is
...
+        // get deposit bond
+        IBondController depositBond = perp.getDepositBond();
+        BondTranches memory bt = depositBond.getTranches();
+
+        // NOTE: the vault may already have seniors which
+        //       we have to ignore for the sake of this operation.
+        uint256 seniorBalBefore = bt.tranches[0].balanceOf(address(this));
+
+        // transfer underlying in
+        underlying.safeTransferFrom(_msgSender(), address(this), amount);
```
aalavandhan, line 428: `perp.deposit` which comes after compares vault and spot TVL which could be potentially manipulated by this action. 

I'm rethinking if this function needs to be part of the vault or some other abstraction which sits above {perp,vault}

spot-contracts/contracts/vaults/RolloverVault.sol
```diff
@@ -404,6 +404,58 @@ contract RolloverVault is
...
+            _syncAndAddDeployedAsset(bt.tranches[i]);
+        }
+
+        // transfer perps back to user
+        IERC20Upgradeable(address(perp)).safeTransfer(_msgSender(), perpsMinted);
+        _syncAsset(perp);
+
+        // mint vault notes using remaning junior tranches
+        uint256 valueIn = getTVL() - tvlBefore;
+        uint256 notesMinted = (totalSupply_ > 0) ? totalSupply_.mulDiv(valueIn, tvlBefore) : (valueIn * INITIAL_RATE);
```
nms-7, line 450: should we move this to a helper function or pricing strategy, if we ever want to change how we price z's for minting?

PR #186, opened 2023-12-11
Monetary policy for Spot

aalavandhan:
* Fee policy controls all fees for both `SPOT` and `stAMPL`
* Perp uses the new fee policy
* [Vault code refactor](https://github.com/ampleforth/spot/pull/183)
* Vault v2 storage updates, with mint redeem and deployment fees


spot-contracts/contracts/FeePolicy.sol
```diff
@@ -0,0 +1,259 @@
...
+    function computeSubscriptionState(IBondIssuer bondIssuer) public returns (SubscriptionState memory) {
+        SubscriptionState memory s;
+        s.currentPerpTVL = perp.getTVL();
+        s.currentVaultTVL = vault.getTVL();
+        // NOTE: We assume that perp minting bonds have only 2 tranches and perp assumes the senior one.
+        //       We assume that perp's TVL and vault's TVL values have the same base denomination.
+        s.targetVaultTVL = s
+            .currentPerpTVL
+            .mulDiv(bondIssuer.trancheRatios(1), bondIssuer.trancheRatios(0), MathUpgradeable.Rounding.Up)
+            .mulDiv(ONE, targetSubscriptionRatio, MathUpgradeable.Rounding.Up);
```
aalavandhan, line 255:
```suggestion
            .mulDiv(targetSubscriptionRatio, ONE, MathUpgradeable.Rounding.Up);
```

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -0,0 +1,259 @@
...
+    /// @notice Contract initializer.
+    /// @param perp_ Reference to perp.
+    /// @param vault_ Reference to the rollover vault.
+    function init(IPerpetualTranche perp_, IVault vault_) public initializer {
+        __Ownable_init();
+
+        perp = perp_;
+        vault = vault_;
+
+        maxPerpMintFeePerc = ONE / 30; // 0.033
```
brandoniles, line 127: Should we just init everything to 0? I assume whatever these are will be explicitly updated after deployment anyway.

I like having some non-zero values for the curve params, though

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -0,0 +1,259 @@
...
+ *
+ *  @notice This contract determines fees for interacting with the perp and vault systems.
+ *
+ *          The fee policy attempts to balance the demand for holding perp tokens with
+ *          the demand for holding vault tokens; such that the total collateral in the vault
+ *          supports rolling over all mature collateral backing perps.
+ *
+ *          Fees are computed based on the following variables.
+ *              - `currentVaultTVL`  : Total value of collateral in the rollover vault at a given time.
+ *              - `targetVaultTVL`   : The minimum value of collateral that needs to be in the vault
```
brandoniles, line 25: Might be worth talking about naming, since I've had to reference this a bunch to times to keep straight.

The targetTVL here is not actually "the target" as far as behavior goes. The target the system tries to meet is targetSubscriptionRatio * targetTVL.

What about "bondEquilibriumTVL" being what matches the bond ratios
Then target is what's left after the safety buffer is applied?

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -0,0 +1,259 @@
...
+            s.normalizedSubscriptionRatio.toInt256(),
+            perpRolloverFeeAPR.lower,
+            perpRolloverFeeAPR.upper,
+            perpRolloverFeeAPR.growth,
+            ONE.toInt256()
+        );
+
+        // We calculate the rollover fee for the given cycle by dividing the annualized rate
+        // by the number of cycles in any given year.
+        return (rolloverAPR * bondIssuer.maxMaturityDuration().toInt256()) / ONE_YEAR_SEC;
```
brandoniles, line 215: This should be issue frequency, not duration, right?

Dividing by 52 vs dividing by 12 is a big difference

spot-contracts/contracts/FeePolicy.sol
brandoniles: What does this line do? I understand up until here. Naming updates should make this much easier to follow.

spot-contracts/contracts/_interfaces/IFeePolicy.sol
```diff
@@ -0,0 +1,41 @@
...
+
+    /// @notice The percentage of the burnt vault note amount to be withheld as fees,
+    ///         as a fixed-point number with {PRICE_DECIMAL} decimal places.
+    function computeVaultBurnFeePerc() external returns (uint256);
+
+    /// @notice The fixed amount fee withheld by the vault during each deployment,
+    ///         denominated in the underlying collateral asset.
+    function computeVaultDeploymentFee() external returns (uint256);
+
+    /// @notice Number of decimals representing 1.0 or 100%.
```
brandoniles, line 39:
```suggestion
    /// @notice Number of decimals representing a multiplier of 1.0. So, 100% = 1*10**decimals
```
Is this clearer?

brandoniles (commented):
Still need to look at RolloverVault.sol

Looks good so far, but let's regroup on the terminology used in the fee strategy

spot-contracts/contracts/_interfaces/IFeePolicy.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.0;
+
+import { IERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
+import { IBondController } from "./buttonwood/IBondController.sol";
+
+interface IFeePolicy {
+    /// @notice The percentage of the mint perp tokens to be withheld as fees,
```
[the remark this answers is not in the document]
aalavandhan: I think `@notice` and `@return` are standard

https://docs.soliditylang.org/en/v0.8.11/natspec-format.html

spot-contracts/contracts/_interfaces/IFeePolicy.sol
```diff
@@ -0,0 +1,41 @@
...
+    ///         as a fixed-point number with {PRICE_DECIMAL} decimal places.
+    function computePerpMintFeePerc() external returns (uint256);
+
+    /// @notice The percentage of the burnt perp tokens to be withheld as fees,
```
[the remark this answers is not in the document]
aalavandhan: Right. Fees are always withheld from the token(s) going out ..

spot-contracts/contracts/_interfaces/IFeePolicy.sol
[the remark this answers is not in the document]
aalavandhan: In this case, 1e6 should mean 100% or 1.0

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -0,0 +1,259 @@
...
+    /// @notice Reference to the perpetual token.
+    IPerpetualTranche public perp;
+
+    /// @notice Reference to the rollover vault.
+    IVault public vault;
+
+    /// @notice The target deviation i.e) the normalization factor.
+    uint256 public targetSubscriptionRatio;
+
+    /// @notice The perp token's mint fee percentage ceiling.
```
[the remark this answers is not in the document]
aalavandhan: Makes sense

spot-contracts/contracts/FeePolicy.sol
aalavandhan: Sure 👍

spot-contracts/contracts/FeePolicy.sol
aalavandhan: It should be divided by 12 as we rollover the entire supply only 12 times

brandoniles: Makes sense. One option maybe worth considering--We could save some gas by not doing any normalization here in code.

I.e. We configure the curve then use it's output directly. But we configure such that the end result is a "yearly rate".

This would save a division, an external call, and a couple reads at the least.

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -79,9 +67,9 @@ contract RolloverVault is
 
     //-------------------------------------------------------------------------
     // Constants
-    uint8 public constant PERC_DECIMALS = 6;
-    uint256 public constant UNIT_PERC = 10**PERC_DECIMALS;
-    uint256 public constant HUNDRED_PERC = 100 * UNIT_PERC;
+    uint8 public constant PERC_DECIMALS = 8;
```
brandoniles, line 70: I think we need a comment here that this is in the "multiplier space", not the "percent space". Are we consistent across the codebase?

Sorry, I've made a bunch on comments on this topic but I can see this causing a mistake at some point.

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -125,51 +108,98 @@ contract RolloverVault is
     /// @notice The perpetual token on which rollovers are performed.
     IPerpetualTranche public perp;
 
+    //--------------------------------------------------------------------------
+    // v2.0.0 STORAGE ADDITION
+
+    /// @notice External contract that controls fees for minting, redeeming, swapping and deployment.
```
brandoniles, line 114: Maybe just say that this orchestrates fees across the spot protocol. If we list specific fees, those can get out of date.

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -125,51 +108,98 @@ contract RolloverVault is
...
+    IFeePolicy public feePolicy;
+
+    /// @notice Reference to the wallet or contract that has the ability to pause/unpause operations.
```
brandoniles, line 117:
```suggestion
    /// @notice Reference to the address that has the ability to pause/unpause operations.
```

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -125,51 +108,98 @@ contract RolloverVault is
...
-    // ADMIN only methods
+        // set the reference to the fee policy
+        updateFeePolicy(feePolicy_);
 
-    /// @notice Pauses deposits, withdrawals and vault operations.
-    /// @dev NOTE: ERC-20 functions, like transfers will always remain operational.
-    function pause() external onlyOwner {
-        _pause();
+        // setting initial parameter values
+        minDeploymentAmt = 0;
```
brandoniles, line 172: Since uninitialized contract values are 0, think we need these?

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -565,33 +484,87 @@ contract RolloverVault is
...
+                _syncAndRemoveDeployedAsset(bt.tranches[j]);
+            }
+        }
+
+        // sync underlying
+        _syncAsset(underlying);
+    }
+
+    /// @dev Redeems perp tokens held by the vault for tranches.
+    function _redeemPerp(IPerpetualTranche perp_) private {
```
brandoniles, line 528: Can this eventually be removed, once all the perps are redeemed and no more come in?

Guess it could still be used in future minting / swapping flows?

spot-contracts/contracts/_interfaces/IFeePolicy.sol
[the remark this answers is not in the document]
aalavandhan: I dont think @return applies in the context of a struct definition

spot-contracts/contracts/_interfaces/IFeePolicy.sol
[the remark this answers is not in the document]
aalavandhan: Sure. I think the updated version has this

spot-contracts/contracts/FeePolicy.sol
aalavandhan: Right. That would save the extra query. We just have to be extra cautious while setting the parameters ..

spot-contracts/contracts/RolloverVault.sol
aalavandhan: Yea there are 2 consumers, perp and the vault. I renamed it to `FEE_POLICY_DECIMALS` and `ONE_FEE_PERC` which means 100% or 1.0

spot-contracts/contracts/RolloverVault.sol
aalavandhan: Not really, but prefer keeping it in there for clarity

spot-contracts/contracts/RolloverVault.sol
aalavandhan: Right this is essential for swapping flows

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -0,0 +1,260 @@
...
+        perpRolloverFee.upper = int256(ONE) / 10; // 0.1
+        perpRolloverFee.growth = 5 * int256(ONE); // 5.0
+
+        targetSubscriptionRatio = (ONE * 133) / 100; // 1.33
+    }
+
+    /// @notice Updates the target subscription ratio.
+    /// @param targetSubscriptionRatio_ The new target subscription ratio as a fixed point number with {DECIMALS} places.
+    function updateTargetSubscriptionRatio(uint256 targetSubscriptionRatio_) external onlyOwner {
+        require(targetSubscriptionRatio_ > 0, "FeeStrategy: invalid subscription");
```
nms-7, line 141: consider upper/lower bounds to prevent decimal errors, e.g.

(ONE * 50) / 100 < targetSubscriptionRatio < ONE * 2

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -0,0 +1,260 @@
...
+    /// @inheritdoc IFeePolicy
+    function computePerpBurnFeePerc() external override returns (uint256) {
+        return isOverSubscribed() ? perpBurnFeePerc : 0;
+    }
+
+    /// @inheritdoc IFeePolicy
+    function computePerpRolloverFeePerc() external override returns (int256) {
+        SubscriptionState memory s = computeSubscriptionState();
+
+        // We calculate the rollover fee for the given cycle by dividing the annualized rate
```
nms-7, line 202: obsolete comment?
```suggestion
```

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -0,0 +1,260 @@
...
+        // by the number of cycles in any given year.
```
nms-7, line 203:
```suggestion
```

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -177,8 +149,8 @@ contract PerpetualTranche is
     uint8 public constant PRICE_DECIMALS = 8;
     uint256 public constant UNIT_PRICE = (10**PRICE_DECIMALS);
 
-    uint8 public constant PERC_DECIMALS = 8;
-    uint256 public constant HUNDRED_PERC = 10**PERC_DECIMALS;
+    uint8 public constant FEE_POLICY_DECIMALS = 8;
+    uint256 public constant FEE_ONE_PERC = (10**FEE_POLICY_DECIMALS);
```
nms-7, line 153: naming is a bit awkward. maybe `FEE_UNIT`

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -79,9 +67,8 @@ contract RolloverVault is
 
     //-------------------------------------------------------------------------
     // Constants
-    uint8 public constant PERC_DECIMALS = 6;
-    uint256 public constant UNIT_PERC = 10**PERC_DECIMALS;
-    uint256 public constant HUNDRED_PERC = 100 * UNIT_PERC;
+    uint8 public constant FEE_POLICY_DECIMALS = 8;
+    uint256 public constant FEE_ONE_PERC = (10**FEE_POLICY_DECIMALS);
```
nms-7, line 71: same here re: naming

spot-contracts/contracts/RolloverVault.sol
brandoniles: I think that's clear 👍

spot-contracts/contracts/PerpetualTranche.sol
brandoniles: I like how the existing version explicitly says "One percent", given the previous discussions.

Because "unit" could mean "1.0 as in 1%", but it could also mean "1.0 as in 100%"

brandoniles (approved):
LGTM, but also get's Nick's review

spot-contracts/contracts/PerpetualTranche.sol
aalavandhan: Yes prefer `FEE_ONE_PERC`. Open better options though ..

nms-7 (approved):
LGTM

PR #187, opened 2023-12-18
Vault melding v2

aalavandhan:
* Vault now allows for 2way swapping between perps and the underlying
* Fee policy controls swap fees and limits

### DIFF
--- a/spot-contracts/contracts/FeePolicy.sol
+++ b/spot-contracts/contracts/FeePolicy.sol
@@ -132,7 +132,7 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
         vaultBurnFeePerc = 0;
 
         // NOTE: With the current bond length of 28 days, rollover rate is annualized by dividing by: 365/28 ~= 13
-        perpRolloverFee.lower = -int256(ONE) / (300 * 13); // -0.033/13 = -0.00253 (3.3% annualized)
+        perpRolloverFee.lower = -int256(ONE) / (30 * 13); // -0.033/13 = -0.00253 (3.3% annualized)
         perpRolloverFee.upper = int256(ONE) / (10 * 13); // 0.1/13 = 0.00769 (10% annualized)
         perpRolloverFee.growth = 5 * int256(ONE); // 5.0
 
@@ -205,7 +205,7 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
         address assetIn,
         address assetOut,
         uint256 feePerc
-    ) public onlyOwner {
+    ) external onlyOwner {
         require(feePerc <= ONE, "FeeStrategy: perc too high");
         bytes32 pairHash = keccak256(abi.encodePacked(assetIn, assetOut));
         if (feePerc > 0) {
@@ -231,10 +231,20 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
     }
 
     /// @inheritdoc IFeePolicy
-    function computePerpBurnFeePerc() external override returns (uint256) {
-        IFeePolicy.SubscriptionState memory s = computeSubscriptionState();
+    function computePerpBurnFeePerc(uint256 perpAmtBurnt, uint256 perpTotalSupply) external override returns (uint256) {
+        // The act of burning perps increases the subscription ratio,
+        // We thus have to check if the "post"-burning subscription state to account for fees.
+
+        // NOTE: The perp and vault TVLs are denominated in the underlying asset.
+        // We calulate the perp post-burn TVL, by multiplying the current tvl by
+        // the fraction of supply remaning.
+        IFeePolicy.SubscriptionState memory postBurnState = computeSubscriptionState(
+            perp.getDepositBond(),
+            perp.getTVL().mulDiv(perpTotalSupply - perpAmtBurnt, perpTotalSupply),
+            vault.getTVL()
+        );
         // When the system is over-subscribed there exists an active redemption fee
-        return (s.normalizedSubscriptionRatio > 1) ? perpBurnFeePerc : 0;
+        return (postBurnState.normalizedSubscriptionRatio > 1) ? perpBurnFeePerc : 0;
     }
 
     /// @inheritdoc IFeePolicy
@@ -268,7 +278,7 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
     /// @inheritdoc IFeePolicy
     function computeUnderlyingToPerpSwapFeePercs(uint256 valueIn) external override returns (uint256, uint256) {
         // When user swaps underlying for vault's perps -> perps are minted by the vault
-        // Similar to perp mint fees, here too we use the "post"-minting subscription state.
+        // Similar to perp mint fees, here too check the "post"-minting subscription state.
         uint256 currentPerpTVL = perp.getTVL();
         uint256 currentVaultTVL = vault.getTVL();
         IFeePolicy.SubscriptionState memory postMintState = computeSubscriptionState(
@@ -290,20 +300,27 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
     }
 
     /// @inheritdoc IFeePolicy
-    function computePerpToUnderlyingSwapFeePercs() external override returns (uint256, uint256) {
+    function computePerpToUnderlyingSwapFeePercs(uint256 valueIn) external override returns (uint256, uint256) {
         // When user swaps perps for vault's underlying -> perps are redeemed by the vault
-        IFeePolicy.SubscriptionState memory s = computeSubscriptionState();
+        // Similar to perp burn fees, here too check the "post"-burn subscription state.
+        uint256 currentPerpTVL = perp.getTVL();
+        uint256 currentVaultTVL = vault.getTVL();
+        IFeePolicy.SubscriptionState memory postBurnState = computeSubscriptionState(
+            perp.getDepositBond(),
+            currentPerpTVL - valueIn,
+            currentVaultTVL
+        );
 
         // Split the fees between the perp and vault systems
         uint256 vaultFeePerc = vaultSwapFeePerc[
             keccak256(abi.encodePacked(address(perp), address(vault.underlying())))
         ];
-        // When the system is over-subscribed, we charge no perp burn fee.
-        if (s.normalizedSubscriptionRatio > 1) {
+        // When the system is under-subscribed, we charge no perp burn fee.
+        if (postBurnState.normalizedSubscriptionRatio <= 1) {
             return (0, vaultFeePerc);
         }
 
-        // When the system is under-subscribed, we charge a perp burn fee on top of the vault's swap fee.
+        // When the system is over-subscribed, we charge a perp burn fee on top of the vault's swap fee.
         return (perpBurnFeePerc, vaultFeePerc);
     }
 
@@ -318,7 +335,7 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
         return computeSubscriptionState(perp.getDepositBond(), perp.getTVL(), vault.getTVL());
     }
 
-    /// @notice Computes the detailed subscription state based on the recorded `perpTVL` and `vaultTVL`.
+    /// @notice Computes the detailed subscription state based on the given `perpTVL` and `vaultTVL`.
     /// @return The subscription state.
     function computeSubscriptionState(
         IBondController depositBond,
@@ -328,12 +345,12 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
         IFeePolicy.SubscriptionState memory s;
         s.perpTVL = perpTVL;
         s.vaultTVL = vaultTVL;
-        // NOTE: We assume that perp only accepts  assumes the senior one.
+        // NOTE: We assume that perp only accepts the senior one.
         //       We assume that perp's TVL and vault's TVL values have the same base denomination.
         (, uint256 perpTR) = depositBond.tranches(0);
         uint256 equilibriumVaultTVL = s.perpTVL.mulDiv(
-            perpTR,
             TRANCHE_RATIO_GRANULARITY - perpTR,
+            perpTR,
             MathUpgradeable.Rounding.Up
         );
         s.normalizedSubscriptionRatio = s.vaultTVL.mulDiv(ONE, equilibriumVaultTVL).mulDiv(
--- a/spot-contracts/contracts/FeePolicy.sol
+++ b/spot-contracts/contracts/FeePolicy.sol
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.19;
 
 import { IFeePolicy, IERC20Upgradeable } from "./_interfaces/IFeePolicy.sol";
-import { IPerpetualTranche, IBondIssuer } from "./_interfaces/IPerpetualTranche.sol";
+import { IPerpetualTranche, IBondController } from "./_interfaces/IPerpetualTranche.sol";
 import { IVault } from "./_interfaces/IVault.sol";
 
 import { MathUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/math/MathUpgradeable.sol";
@@ -10,6 +10,8 @@ import { SafeCastUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/m
 import { AddressUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/AddressUpgradeable.sol";
 import { OwnableUpgradeable } from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 import { Sigmoid } from "./_utils/Sigmoid.sol";
+
+import { UnacceptableSwap } from "./_interfaces/ProtocolErrors.sol";
 
 /**
  *  @title FeePolicy
@@ -21,14 +23,13 @@ import { Sigmoid } from "./_utils/Sigmoid.sol";
  *          supports rolling over all mature collateral backing perps.
  *
  *          Fees are computed based on the following variables.
- *              - `currentVaultTVL`     : Total value of collateral in the rollover vault at a given time.
- *              - `equilibriumVaultTVL` : The minimum value of collateral that needs to be in the vault
- *                                        to sustain rolling over the entire perp supply.
- *                                        It is the current tvl of perp divided by the tranche ratio of seniors used to mint perp.
- *              - `currentSubscriptionRatio`    : The ratio between the `currentVaultTVL` and the `equilibriumVaultTVL`.
- *              - `targetSubscriptionRatio`     : The subscription ratio above which the system is considered "over-subscribed".
+ *              - `vaultTVL`                    : Total value of collateral in the rollover vault at a given time.
+ *              - `equilibriumVaultTVL`         : The minimum value of collateral that needs to be in the vault
+ *                                                to sustain rolling over the entire perp supply.
+ *              - `subscriptionRatio`           : The ratio between the `vaultTVL` and the `equilibriumVaultTVL`.
+ *              - `targetSubscriptionRatio`     : The ratio above which the system is considered "over-subscribed".
  *                                                Adds a safety buffer to ensure that rollovers are better sustained.
- *              - `normalizedSubscriptionRatio` : The ratio between `currentSubscriptionRatio` and the `targetSubscriptionRatio`.
+ *              - `normalizedSubscriptionRatio` : The ratio between `subscriptionRatio` and the `targetSubscriptionRatio`.
  *
  *          When the system is "under-subscribed":
  *              - Rollover fees flow from perp holders to vault note holders.
@@ -43,12 +44,16 @@ import { Sigmoid } from "./_utils/Sigmoid.sol";
  *          Regardless of the subscription ratio, the system charges a fixed percentage fee
  *          for minting and redeeming vault notes.
  *
+ *          The system favors an elastic perp supply and an inelastic vault note supply.
+ *
  *          The rollover fees are signed and can flow in either direction based on the subscription ratio.
  *          The fee is a percentage is computed through a sigmoid function.
  *          The slope and asymptotes are set by the owner.
  *          CRITICAL: The rollover fee percentage is NOT annualized, the fee percentage is applied per rollover.
  *          The number of rollovers per year changes based on the duration of perp's minting bond.
  *
+ *          We consider a `normalizedSubscriptionRatio` of greater than 1.0 healthy.
+ *          Minting additional perps or redeeming vault notes reduces the subscription ratio.
  *
  */
 contract FeePolicy is IFeePolicy, OwnableUpgradeable {
@@ -57,10 +62,14 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
     using SafeCastUpgradeable for uint256;
 
     /// @dev The returned fee percentages are fixed point numbers with {DECIMALS} places.
-    ///      The decimals should line up with value expected by consumer (perp).
+    ///      The decimals should line up with value expected by consumer (perp, vault).
     ///      NOTE: 10**DECIMALS => 100% or 1.0
     uint8 public constant DECIMALS = 8;
     uint256 public constant ONE = (1 * 10**DECIMALS); // 1.0 or 100%
+
+    /// @dev Using the same granularity as the underlying buttonwood tranche contracts.
+    ///      https://github.com/buttonwood-protocol/tranche/blob/main/contracts/BondController.sol
+    uint256 private constant TRANCHE_RATIO_GRANULARITY = 1000;
 
     uint256 public constant SIGMOID_BOUND = ONE / 100; // 0.01 or 1%
     uint256 public constant SR_LOWER_BOUND = (ONE * 75) / 100; // 0.75 or 75%
@@ -92,6 +101,9 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
     ///      Paid by the vault note holders to the system owner.
     uint256 public vaultDeploymentFee;
 
+    /// @notice Mapping between `hash({assetIn,assetOut})` and the fee percentage.
+    mapping(bytes32 => uint256) public vaultSwapFeePerc;
+
     struct RolloverFeeSigmoidParams {
         /// @notice Lower asymptote
         int256 lower;
@@ -103,18 +115,6 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
 
     /// @notice Parameters which control the asymptotes and the slope of the perp token's rollover fee.
     RolloverFeeSigmoidParams public perpRolloverFee;
-
-    /// @dev The current subscription state the vault system relative to the perp supply.
-    struct SubscriptionState {
-        /// @notice The current tvl of perp.
-        uint256 currentPerpTVL;
-        /// @notice The current tvl of the rollover vault.
-        uint256 currentVaultTVL;
-        /// @notice The equilibrium tvl of the rollover vault.
-        uint256 equilibriumVaultTVL;
-        /// @notice Computed normalized subscription ratio.
-        uint256 normalizedSubscriptionRatio;
-    }
 
     /// @notice Contract initializer.
     /// @param perp_ Reference to perp.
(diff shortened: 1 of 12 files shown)


spot-contracts/contracts/RolloverVault.sol
```diff
@@ -542,6 +645,44 @@ contract RolloverVault is
...
+            uint256 underlyingAmtToTranche = depositBond.estimateDepositAmt(seniorsRequired, bt.trancheRatios[0]);
+            _tranche(depositBond, bt, underlyingAmtToTranche);
+        }
+
+        // mint perps
+        _checkAndApproveMax(tranche, address(perp_), seniorsToMintPerps);
+        perp_.deposit(tranche, seniorsToMintPerps);
+
+        // sync holdings
+        _syncAndAddDeployedAsset(tranche);
```
aalavandhan, line 685:
```suggestion
        _syncAndRemoveDeployedAsset(tranche);
```

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -282,26 +282,35 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
             keccak256(abi.encodePacked(address(perp), address(vault.underlying())))
         ];
-        uint256 perpSharePerc = totalFeePerc.mulDiv(currentPerpTVL, currentPerpTVL + currentVaultTVL);
+        // When the system is over-subscribed, we charge no perp burn fee.
+        // The perp's share of the swap fee is zero and the vault gets the entire slice.
+        if (s.normalizedSubscriptionRatio > 1) {
+            return (0, totalFeePerc);
+        }
+
+        // When the system is under-subscribed, perp's share is computed as the minimum of
```
aalavandhan, line 308: Alternatively, we could give perp  `perpBurnFeePerc` and the vault `totalFeePerc-perpBurnFeePerc` provided `totalFeePerc >= perpBurnFeePerc`.

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -227,32 +265,78 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
+    /// @inheritdoc IFeePolicy
+    function computePerpToUnderlyingSwapFeePercs() external override returns (uint256, uint256) {
+        // When user swaps perps for vault's underlying -> perps are redeemed by the vault
+        IFeePolicy.SubscriptionState memory s = computeSubscriptionState();
+
+        // Split the fees between the perp and vault systems
+        uint256 vaultFeePerc = vaultSwapFeePerc[
+            keccak256(abi.encodePacked(address(perp), address(vault.underlying())))
+        ];
+        // When the system is over-subscribed, we charge no perp burn fee.
```
brandoniles, line 301: Is this backwards? I thought...
undersubscribed -> no burn fee
oversubscribed -> burn fee

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -227,32 +265,78 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
-        // NOTE: We assume that perp minting bonds have only 2 tranches and perp assumes the senior one.
+    function computeSubscriptionState(
+        IBondController depositBond,
+        uint256 perpTVL,
+        uint256 vaultTVL
+    ) public view returns (IFeePolicy.SubscriptionState memory) {
+        IFeePolicy.SubscriptionState memory s;
+        s.perpTVL = perpTVL;
+        s.vaultTVL = vaultTVL;
+        // NOTE: We assume that perp only accepts  assumes the senior one.
```
brandoniles, line 331:
```suggestion
        // NOTE: We assume that perp only accepts the senior one.
```

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -227,32 +265,78 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
         //       We assume that perp's TVL and vault's TVL values have the same base denomination.
-        s.equilibriumVaultTVL = s.currentPerpTVL.mulDiv(
-            bondIssuer.trancheRatios(1),
-            bondIssuer.trancheRatios(0),
+        (, uint256 perpTR) = depositBond.tranches(0);
+        uint256 equilibriumVaultTVL = s.perpTVL.mulDiv(
```
brandoniles, line 334: Say perp TVL is 100 and the bond is [25/75]

eq vault TVL should be 300:
= 100 * (750/250)

so, shouldn't this be:
`s.perpTVL.mulDiv(TRANCHE_RATIO_GRANULARITY - perpTR, perpTR)` ?

brandoniles (commented):
Covers everything except RolloverVault.sol

I like where we landed for the fee breakdown strategy

spot-contracts/contracts/_utils/BondHelpers.sol
```diff
@@ -74,6 +74,31 @@ library BondHelpers {
...
+    /// @param mintingTrancheRatio The tranche ratio of the minting tranche.
+    /// @return The number of collateral tokens to deposit.
+    function estimateDepositAmt(
+        IBondController b,
+        uint256 trancheAmtToMint,
+        uint256 mintingTrancheRatio
+    ) internal view returns (uint256) {
+        return
+            trancheAmtToMint
+                .mulDiv(BPS, BPS - b.feeBps(), MathUpgradeable.Rounding.Up)
```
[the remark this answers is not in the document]
aalavandhan: Sounds good

spot-contracts/contracts/_utils/BondHelpers.sol
```diff
@@ -74,6 +74,31 @@ library BondHelpers {
...
     }
 
+    /// @notice The function estimates the amount of collateral which needs to be tranched,
+    ///         to mint the given number of tranche tokens with the provided tranche ratio.
+    /// @dev This function is NOT a precise estimate, ie) when you tranche the estimated amount
+    ///      of collateral you might end up minting slightly more than `trancheAmtToMint`.
+    ///      It is the inverse of the `previewDeposit` function.
+    /// @param b The address of the bond contract.
+    /// @param trancheAmtToMint The amount of tranche tokens to mint.
+    /// @param mintingTrancheRatio The tranche ratio of the minting tranche.
```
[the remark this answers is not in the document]
aalavandhan: It's the ratio of the tranche we'd like to mint. For example say we want to mint 100 Zs and would like to estimate the amount of AMPL required, we'd feed in the tranche ratio of the Z tranche.

spot-contracts/contracts/_interfaces/IFeePolicy.sol
```diff
@@ -5,37 +5,60 @@ import { IERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC
...
-    /// @notice The fixed amount fee charged by the vault during each deployment,
+    /// @return The fixed amount fee charged by the vault during each deployment,
     ///         denominated in the underlying collateral asset.
     function computeVaultDeploymentFee() external returns (uint256);
 
-    /// @notice Number of decimals representing a multiplier of 1.0. So, 100% = 1*10**decimals
+    /// @param valueIn The value of underlying tokens swapped in.
+    /// @return The percentage of perp tokens out to be charged as swap fee by the perp and vault systems,
+    ///         as a fixed-point numbers with {DECIMALS} decimal places.
+    function computeUnderlyingToPerpSwapFeePercs(uint256 valueIn) external returns (uint256, uint256);
```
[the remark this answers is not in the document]
aalavandhan: Sure will do it in a later PR

spot-contracts/contracts/_interfaces/IFeePolicy.sol
```diff
@@ -5,37 +5,60 @@ import { IERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC
...
+
+    /// @return The percentage of underlying tokens out to be charged as swap fee by the perp and vault systems,
+    ///         as a fixed-point numbers with {DECIMALS} decimal places.
+    function computePerpToUnderlyingSwapFeePercs() external returns (uint256, uint256);
+
+    /// @return Number of decimals representing 1.0 or 100%.
```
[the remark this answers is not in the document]
aalavandhan: How about

```
/// @return The number of decimal places used by policy computed fee percentages. (10**decimals()) => 100% or 1.0
```

spot-contracts/contracts/_interfaces/IFeePolicy.sol
```diff
@@ -5,37 +5,60 @@ import { IERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC
...
     function computeVaultBurnFeePerc() external returns (uint256);
 
-    /// @notice The fixed amount fee charged by the vault during each deployment,
+    /// @return The fixed amount fee charged by the vault during each deployment,
     ///         denominated in the underlying collateral asset.
     function computeVaultDeploymentFee() external returns (uint256);
 
-    /// @notice Number of decimals representing a multiplier of 1.0. So, 100% = 1*10**decimals
+    /// @param valueIn The value of underlying tokens swapped in.
+    /// @return The percentage of perp tokens out to be charged as swap fee by the perp and vault systems,
```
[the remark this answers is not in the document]
aalavandhan: Fixed

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -230,11 +229,10 @@ contract PerpetualTranche is
...
 
-    /// @notice Set of all authorized addresses which can execute rollovers.
-    /// @dev The contract owner can modify this set.
-    ///      NOTE: If the set is empty, all addresses are considered authorized and can execute rollovers.
-    ///            else only addresses in the set can execute rollovers.
-    EnumerableSetUpgradeable.AddressSet private _rollers;
+    /// @notice Address of the authorized rollover vault.
+    /// @dev If this address is set, only the rollover vault can perform rollovers.
+    ///      If not rollovers are publically accessible.
+    address public vault;
```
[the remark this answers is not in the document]
aalavandhan: Right, we haven't deployed the whitelisted rollers change yet.

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -131,8 +131,9 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
         vaultMintFeePerc = 0;
         vaultBurnFeePerc = 0;
 
-        perpRolloverFee.lower = -int256(ONE) / 30; // -0.033
-        perpRolloverFee.upper = int256(ONE) / 10; // 0.1
+        // NOTE: With the current bond length of 28 days, rollover rate is annualized by dividing by: 365/28 ~= 13
+        perpRolloverFee.lower = -int256(ONE) / (300 * 13); // -0.033/13 = -0.00253 (3.3% annualized)
```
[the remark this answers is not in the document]
aalavandhan: Good catch. It's `30`

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -178,30 +190,56 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
+        // When the vault is under-subscribed there exists an active mint fee
+        return (postMintState.normalizedSubscriptionRatio <= 1) ? perpMintFeePerc : 0;
     }
 
     /// @inheritdoc IFeePolicy
     function computePerpBurnFeePerc() external override returns (uint256) {
-        return isOverSubscribed() ? perpBurnFeePerc : 0;
+        IFeePolicy.SubscriptionState memory s = computeSubscriptionState();
+        // When the vault is over-subscribed there exists an active redemption fee
+        return (s.normalizedSubscriptionRatio > 1) ? perpBurnFeePerc : 0;
```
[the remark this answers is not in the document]
aalavandhan: It makes sense to check the post op subscription state in the case of perp minting and vault redemption. As these operations will push the system away from the tranche ratio. On the other hand, we know that perp redemption and vault minting push to the system toward the tranche ratio and thus we do not have to check the state.

We currently do not check the post op state after vault redemption because, we charge a fee regardless of the system state. However eventually, I can see us moving toward state dependent vault redemption fee.

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -227,32 +265,76 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
+        uint256 currentPerpTVL = perp.getTVL();
+        uint256 currentVaultTVL = vault.getTVL();
+        IFeePolicy.SubscriptionState memory postMintState = computeSubscriptionState(
+            perp.getDepositBond(),
+            currentPerpTVL + valueIn,
+            currentVaultTVL
+        );
+
+        // If minting leaves the vault under-subscribed, swapping by minting perps is NOT allowed.
+        if (postMintState.normalizedSubscriptionRatio <= 1) {
```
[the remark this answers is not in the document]
aalavandhan: With the current implementation, it will explicitly revert (and not keep the vault assets). We revert when zero assets are swapped in or out.

spot-contracts/contracts/FeePolicy.sol
aalavandhan: Good catch, When the system is undersubscribed no perp redemption fee (we make redemptions easy); when the system is oversubscribed we charge a perp redemption fee.

spot-contracts/contracts/FeePolicy.sol
aalavandhan: Good catch, that's right.

aalavandhan:
> Covers everything except RolloverVault.sol
> 
> I like where we landed for the fee breakdown strategy

Addressed them here
https://github.com/ampleforth/spot/pull/187/commits/ecaeea0e73832383ce204c022c5058729a1280d3

spot-contracts/contracts/_utils/BondHelpers.sol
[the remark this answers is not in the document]
brandoniles: Ok, I think my confusion comes from the name having "ratio" in it.

Say the bond config is [A/Z] = [250/750], and we're minting Zs--

mintingTrancheRatio is 750? I.e. just the numerator, not the "ratio" of 750/1000

spot-contracts/contracts/_interfaces/IFeePolicy.sol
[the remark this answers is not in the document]
brandoniles: How about the original language?
```
Number of decimals representing a multiplier of 1.0. So, 100% = 1*10**decimals
```

spot-contracts/contracts/FeePolicy.sol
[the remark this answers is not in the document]
aalavandhan: Ah right. That makes sense.  We should always check the post-op state for all operations (perp mint and redeem, vault mint and redeem). However since the vault fees are not state dependent (are a fixed percentage), we can skip.

aalavandhan:
I've updated the fee policy to check for the post-op state on both perp minting and burning

https://github.com/ampleforth/spot/pull/187/commits/0aced6adcbdeab9e487f314bd0f43bdd10142dab
https://github.com/ampleforth/spot/pull/187#discussion_r1436638453

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -178,30 +190,62 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
 
     /// @inheritdoc IFeePolicy
-    function computePerpBurnFeePerc() external override returns (uint256) {
-        return isOverSubscribed() ? perpBurnFeePerc : 0;
+    function computePerpBurnFeePerc(uint256 perpAmtBurnt, uint256 perpTotalSupply) external override returns (uint256) {
+        // The act of burning perps increases the subscription ratio,
+        // We thus have to check if the "post"-burning subscription state to account for fees.
+        IFeePolicy.SubscriptionState memory postBurnState = computeSubscriptionState(
+            perp.getDepositBond(),
+            perp.getTVL().mulDiv(perpTotalSupply - perpAmtBurnt, perpTotalSupply),
```
brandoniles, line 239: I would have expected `perp.getTVL() - perpAmtBurnt`. Is there a reason for the mulDiv?

aalavandhan: TVL is denominated in the underlying asset. So we'd have to do `perp.getTVL() - (perpAmtBurnt*perp.getAvgPrice)` ..

brandoniles: Ah makes sense. I'd add a comment, since that's important.

aalavandhan: Sure added.

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -394,6 +400,118 @@ contract RolloverVault is
         return redemptions;
     }
 
+    /// @notice Allows users to swap their underlying tokens for perps held by the vault
+    /// @param underlyingAmtIn The amount of underlying tokens swapped in.
+    /// @param execRecover If set to true, recovers vault assets to maximize liquidity for swapping.
+    /// @return The amount of perp tokens swapped out.
+    function swapUnderlyingForPerps(uint256 underlyingAmtIn, bool execRecover)
+        public
```
brandoniles, line 408: external?

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -394,6 +400,118 @@ contract RolloverVault is
...
+        uint256 perpPrice = perp.getAvgPrice();
+        uint256 perpAmtOut = underlyingAmtIn.mulDiv(PERP_UNIT_PRICE, perpPrice);
+
+        // Compute fees
+        (uint256 swapFeePerpSharePerc, uint256 swapFeeVaultSharePerc) = feePolicy.computeUnderlyingToPerpSwapFeePercs(
+            underlyingAmtIn
+        );
+
+        // We compute perp's fee share
+        uint256 perpAmtToBurn = perpAmtOut.mulDiv(swapFeePerpSharePerc, FEE_ONE_PERC, MathUpgradeable.Rounding.Up);
```
brandoniles, line 423: This is another tricky area with 1 vs 100 and decimals...

I'm always way more comfortable dividing by a number denominated in DECIMALS, rather than a scalar value. Because functionally, we're dividing in order to "fix decimals" not to affect the "numbers".

Commenting here, not because it's wrong per se, but it's definitely a code smell suggesting that we've organized our concepts incorrectly.

I forget now, are our decimals for 100 or for 1? Can we divide by decimals here without having to, e.g. subtract 2 from it?

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -394,6 +400,118 @@ contract RolloverVault is
...
+
+        // We deduct fees by transferring out fewer perp tokens
```
brandoniles, line 425: There are two options: we could transfer out fewer perps, or we could withhold AMPL before minting perps to begin with.

I think the second one would require less gas and seems easier, wdyt? Also, potentially leaves the vault with fewer perps that it'll probably end up reclaiming anyway. The vault generally prefers to hold AMPL over perps.

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -394,6 +400,118 @@ contract RolloverVault is
...
+
+        return perpAmtOut;
+    }
+
+    /// @notice Allows users to swap their perp tokens for underlying tokens held by the vault
+    /// @param perpAmtIn The amount of perp tokens swapped in.
+    /// @param execRecover If set to true, recovers vault assets to maximize liquidity for swapping.
+    /// @return The amount of underlying tokens swapped out.
+    function swapPerpsForUnderlying(uint256 perpAmtIn, bool execRecover)
+        public
```
brandoniles, line 464: external?

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -394,6 +400,118 @@ contract RolloverVault is
         return redemptions;
     }
 
+    /// @notice Allows users to swap their underlying tokens for perps held by the vault
+    /// @param underlyingAmtIn The amount of underlying tokens swapped in.
+    /// @param execRecover If set to true, recovers vault assets to maximize liquidity for swapping.
+    /// @return The amount of perp tokens swapped out.
+    function swapUnderlyingForPerps(uint256 underlyingAmtIn, bool execRecover)
```
brandoniles, line 407: I can also imagine a design that doesn't have an `execRecover` input, but instead will lazily execute recover, if it's needed.

Is the benefit simply that it's cheaper to check execRecover from calldata/mem vs checking token balance? It does complicate the interface.

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -542,6 +661,56 @@ contract RolloverVault is
...
+        // Get the minting bond data
+        IBondController depositBond = perp.getDepositBond();
+        BondTranches memory bt = depositBond.getTranches();
+
+        // NOTE: we use only the most senior tranche to mint perps
+        ITranche tranche = bt.tranches[0];
+
+        // Compute the senior tranche amount need to be deposited in order to mint perps.
+        uint256 seniorAmtToDeposit = perpAmtToMint.mulDiv(
+            perpPrice,
```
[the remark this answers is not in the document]
aalavandhan: perp.getAvgPrice() -> returns the balance weighted price of all the tranches in the reserve
perp.computePrice(tranche) -> returns the price of a given tranche

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -542,6 +661,56 @@ contract RolloverVault is
         }
     }
 
+    /// @dev Tranches the vault's underlying to mint perps as needed.
+    ///      If the vault already holds required perps, it skips minting new ones.
+    ///      Additionally, performs some book-keeping to keep track of the vault's assets.
+    function _trancheAndMintPerps(
```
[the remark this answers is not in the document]
aalavandhan: Right I plan to add preview functions in the next pr

spot-contracts/contracts/RolloverVault.sol
aalavandhan: FEE_ONE_PERC = 100% or 1.0

spot-contracts/contracts/RolloverVault.sol
aalavandhan: Gas-wise I don't think it should make any difference as it will involve the same number of operations.

It becomes tricky as there are 2 parts to the fee. The vault share which could be held in either ampl or perps and the perp share which can only be paid by burning perps. It seems cleaner to have both parts be paid in the same token, if possible.

spot-contracts/contracts/RolloverVault.sol
aalavandhan: Checking if recover is executable will require as many reads as actually executing the recover. I imagine most interfaces will invoke this function with a default value of `execRecover=false`. 

Agree we can get rid of this.

spot-contracts/contracts/RolloverVault.sol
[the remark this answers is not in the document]
aalavandhan: But here we could use the vault's `_computeVaultTrancheValue` method rather than `perp.computePrice`, if we are okay assuming both contracts implement the same pricing strategy. However, Asking perp for how the tranche is priced is guaranteed to be correct.

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -433,13 +427,15 @@ contract RolloverVault is
...
-            recover();
-        }
-
         // tranche and mint perps as needed
-        _trancheAndMintPerps(perp, perpPrice, perpAmtOut + perpAmtToBurn);
+        {
+            uint256 perpBal = perp.balanceOf(address(this));
+            uint256 perpsReq = perpAmtOut + perpAmtToBurn;
+            if (perpBal < perpsReq) {
+                recover();
```
brandoniles, line 435: Is it too inefficient to gate the recover op on whether there's enough underlying balance?

Gating tranching based on whether there's enough perp makes sense, but recover only *really* needs to happen if there's not enough underlying.

Perhaps, the lazy recover() could be moved into the tranche and mint function, where we already know the required values?

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -433,13 +427,15 @@ contract RolloverVault is
...
         underlying.safeTransferFrom(_msgSender(), address(this), underlyingAmtIn);
 
-        if (execRecover) {
-            // redeem tranches to maximize underlying balance for minting
-            recover();
-        }
-
         // tranche and mint perps as needed
-        _trancheAndMintPerps(perp, perpPrice, perpAmtOut + perpAmtToBurn);
+        {
```
brandoniles, line 431: Any technical reason for the extra scope block or purely for style?

In c++ this actually pushes another frame on the stack, which is useful in some cases, but not sure about solidity.

spot-contracts/contracts/RolloverVault.sol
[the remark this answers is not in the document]
brandoniles: This should be moot, if we move to the design with different mint functions on Perp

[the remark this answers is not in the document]
aalavandhan: will plan to move this logic to preview deposit amount into perp

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -685,7 +674,7 @@ contract RolloverVault is
         // NOTE: we use only the most senior tranche to mint perps
         ITranche tranche = bt.tranches[0];
 
-        // Compute the senior tranche amount need to be deposited in order to mint perps.
+        // Compute the senior tranche amount needed to mint perps.
```
aalavandhan, line 677:
```suggestion
        // TODO: move preview deposit amount fn into perp 
        // Compute the senior tranche amount needed to mint perps.
```

spot-contracts/contracts/RolloverVault.sol
brandoniles: Let's try the second one, since it doesn't seem much more complicated. This should save some extra recover calls.

spot-contracts/contracts/RolloverVault.sol
aalavandhan: the compiler imposes a limit on the number of variable definitions within a scoped block

aalavandhan:
Lets assume that the vault does not hold perps at all

spot-contracts/contracts/RolloverVault.sol
aalavandhan: But wait, I think we're already doing the 2nd one?
Say user transfers 100 AMPL and the fees for perp/vault are 2 and 3 % respectively. 

perpAmt to burn is 2 perps, perpAmt to the user is 95. We first transfer 100 AMPL from the user, then mint 97 perps out of which 95 are transferred to the user and 2 are burnt.

Rather, we don't mint 100 perps and keep 3 as the vault fee.

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -497,24 +471,25 @@ contract RolloverVault is
             IERC20Burnable(address(perp)).burn(perpAmtToBurn);
         }
 
-        if (execRecover) {
-            // redeem perps for tranches and recover underlying
-            recover();
-        }
+        // Redeem perps for tranches and execute recover
+        _redeemPerps(perp);
+        recover();
```
brandoniles, line 476: Wdyt about lazy recover?
```
if (underlyingAmtOut < underlying.balanceOf(this)) {
    recover();
}
underlying.transfer(...);
```

Case for the change: saves gas on the whole. Especially in cases where the vault is way overcollateralized as now, lots of these ops would never need recovers.

Case for keeping as-is: this function is mostly called by arbers, and they're already only calling if it's profitable, so just bundle in the positive externalities here. Result is slightly wider price bands, inline with gas costs for recover.

spot-contracts/contracts/RolloverVault.sol
brandoniles: Oh you're right! Yeah this is great 👍

brandoniles (commented):
Just one last comment. I'm really liking where this is ending up.

spot-contracts/contracts/RolloverVault.sol
brandoniles: As you noted offline, the perp redemption causes a lot of pollution to the tranche set. So makes sense for this caller to execute the cleanup. Let's keep as it is 👍

brandoniles (approved):
LGTM

PR #190, opened 2024-01-19
Contract code cleanup and unit tests

aalavandhan:
* Removed all string require errors in favor of solidity custom errors
* Updated fee policy interface with redundant `computeDeviationRatio` methods
* Perp interface cleanup, skipping update state when system is paused
* Vault interface cleanup, caching and dust handling
* Unit tests


spot-contracts/contracts/BondIssuer.sol
```diff
@@ -102,7 +105,10 @@ contract BondIssuer is IBondIssuer, OwnableUpgradeable {
         for (uint8 i = 0; i < trancheRatios_.length; i++) {
             ratioSum += trancheRatios_[i];
         }
-        require(ratioSum == TRANCHE_RATIO_GRANULARITY, "BondIssuer: Invalid tranche ratios");
+
+        if (ratioSum > TRANCHE_RATIO_GRANULARITY) {
+            revert UnacceptableTrancheRatios();
```
brandoniles, line 110: What do you think about parameterizing errors?
i.e.--
`revert SomeErrorName(expectedValue, actualValue);` or
`revert SomeInvalidValueErrorName(value)`

I'll guess that our basic structures involve a revert guarded by a conditional that tests theses values already, so they should be available in most cases.

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -149,33 +153,41 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
     }
 
     /// @notice Update the parameters determining the slope and asymptotes of the sigmoid fee curve.
     /// @param p Lower, Upper and Growth sigmoid paramters are fixed point numbers with {DECIMALS} places.
     function updatePerpRolloverFees(RolloverFeeSigmoidParams calldata p) external onlyOwner {
-        require(p.lower >= -int256(SIGMOID_BOUND), "FeeStrategy: sigmoid lower bound too low");
-        require(p.upper <= int256(SIGMOID_BOUND), "FeeStrategy: sigmoid upper bound too high");
-        require(p.lower <= p.upper, "FeeStrategy: sigmoid asymptotes invalid");
+        // SIGMOID_BOUND is set to 1%, i.e) the rollover fee can be at most 1% on either direction.
+        // If the bond duration is 28 days and 13 rollovers happen per year,
```
brandoniles, line 190: I think any comments related to numbers should be where those numbers are defined. So where SIGMOID_BOUND is defined.

In the logic where they're used, we treat them as givens.

brandoniles (commented):
For the first commit, bccbb8ba64675d6f1cf1a9a830674b7837a585f5

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -306,13 +310,21 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
+    /// @inheritdoc IFeePolicy
+    function computeDeviationRatio(
+        uint256 perpTVL,
+        uint256 vaultTVL,
+        uint256 seniorTR
+    ) public view returns (uint256) {
         // NOTE: We assume that perp's TVL and vault's TVL values have the same base denomination.
         return
-            s.vaultTVL.mulDiv(s.perpTR, (s.perpTVL * s.vaultTR), MathUpgradeable.Rounding.Up).mulDiv(
+            vaultTVL.mulDiv(ONE, targetSubscriptionRatio).mulDiv(
```
brandoniles, line 325: Does this revert to the earlier, more complicated computation?

brandoniles (commented):
For #2, 5c120932c61286f66fba55958dd743ce3d80f2b7

spot-contracts/contracts/BondIssuer.sol
aalavandhan: I don't really see any use to parameterizing errors right now. I could be useful, if we wanted to raise and catch exceptions downstream and do some more advanced handling; but don't see it being the case anytime soon.

spot-contracts/contracts/FeePolicy.sol
aalavandhan: It is essentially the same, we just calculate `vaultTR` (or the juniorTR) instead of passing it down.

aalavandhan: This would be the right version
```
return
            (vaultTVL*ONE)                
                .mulDiv(seniorTR, (perpTVL * (TRANCHE_RATIO_GRANULARITY - seniorTR)), MathUpgradeable.Rounding.Up)
                .mulDiv(ONE, targetSubscriptionRatio, MathUpgradeable.Rounding.Up);

```

aalavandhan: However the current version seems easier to think about

```
juniorTR = (TRANCHE_RATIO_GRANULARITY - seniorTR)

# expected vault tvl 
vaultTVL' = perpTVL
.mulDiv(juniorTR, seniorTR, MathUpgradeable.Rounding.Up)
.mulDiv(targetSubscriptionRatio, ONE, MathUpgradeable.Rounding.Up)

# deviation
deviation = vaultTVL.mulDiv(ONE, vaultTVL')
```

spot-contracts/contracts/FeePolicy.sol
aalavandhan: Right moved

spot-contracts/contracts/BondIssuer.sol
brandoniles: Sure. If there's value, it's in debugging runtime errors most likely.

spot-contracts/contracts/FeePolicy.sol
brandoniles: I'm normally all about self-documenting code, but I think we can make exceptions in cases where there's gas optimization to be gained and there's provable equivalence.

If we formulate it as:
DR = (vaultTVL * seniorTR) / (perpTVL * juniorTR * targetSR)

then I think we only need one mulDiv? (I'm omitting the ONEs for fixing decimals, but you get the idea)

This also matches to documentation we already have at the top of the file pretty closely

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -629,6 +635,11 @@ contract PerpetualTranche is
     ///      This function is to be invoked on all external function entry points which are
     ///      read the reserve storage. This function is intended to be idempotent.
     function updateState() public override {
+        // Skip state update when system is paused.
+        if (paused()) {
```
brandoniles, line 639: Normally we ensure pause state with the `whenNotPaused` modifier. Any reason you deviated?

That modifier reverts, so is it because to keep execution going?

I think it's safer to revert, in case there are paths that keep going, for whatever reason, that expect an updated state but don't have one. That could lead to weird scenarios.

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -313,6 +313,13 @@ contract PerpetualTranche is
     /// @notice Updates the reference to the rollover vault.
     /// @param newVault The address of the new vault.
     function updateVault(address newVault) public onlyOwner {
+        // If set to a non-zero address, we expect vault to be valid and implement `getTVL()`
+        if (newVault != address(0)) {
```
brandoniles, line 317: As discussed offline, we probably don't need this condition since it's called by the owner.

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -790,21 +860,17 @@ contract PerpetualTranche is
         // We compute "price" as the value of a unit token.
         // The perp, tranche tokens and the underlying are denominated as fixed point numbers
         // with the same number of decimals.
+        IERC20Upgradeable underlying_ = _reserveAt(0);
```
brandoniles, line 863: Need the trailing underscore? I normally associate that with contract vars

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -998,8 +1015,8 @@ contract PerpetualTranche is
     /// @dev Queries the current subscription state of the perp and vault systems.
     function _querySubscriptionState() private view returns (IFeePolicy.SubscriptionParams memory s) {
         s.perpTVL = _reserveValue();
-        s.vaultTVL = IVault(vault).getTVL();
-        (s.perpTR, s.vaultTR) = _depositBond.getSeniorJuniorRatios();
+        s.vaultTVL = (address(vault) != address(0)) ? IVault(vault).getTVL() : 0;
```
brandoniles, line 1018: Do we ever expect vault to be 0x0? Worth the gas for this check?

A lot of things might break, like anything dependent on SRs. Seems like we should write the entire system to work with that case, or not worry about it.

We could enforce during updateVault that it's != 0x0

brandoniles (commented):
For #3, ce6017001e83262f9f9b9cc60c7f37646034820e

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -268,6 +274,10 @@ contract RolloverVault is
         if (!_rollover()) {
             revert InsufficientDeployment();
         }
+
+        // Cleanup rollover; In case that there remain excess seniors and juniors after a successful rollover,
+        // we recover back to underlying.
+        _redeemTranche(trancheIntoPerp);
```
brandoniles, line 290: Is this likely to be significant? I would think tranches coming out of perp would be more important.

brandoniles (commented):
For #4, e0b74e509178d8eaa7ea4530553a518d392c1a5f

spot-contracts/contracts/PerpetualTranche.sol
aalavandhan: We use `whenNotPaused afterStateUpdate` modifiers for deposit, redeem and rollover methods and the `afterStateUpdate` modifier for a bunch of lazy exec read methods. If the `updateState` method reverts, it will brick all of them. I think we still want those methods to work when paused.

spot-contracts/contracts/PerpetualTranche.sol
aalavandhan: updated

spot-contracts/contracts/PerpetualTranche.sol
aalavandhan: We have already method called `underlying` which causes a conflict. Prepending `_` usually indicates private variables or methods.

spot-contracts/contracts/PerpetualTranche.sol
aalavandhan: The `onlyVault` modifier allows all callers to rollover when the vault address (is not set) or set to address(0). However with this current version, the fee policy expects a valid vault reference to compute rollover fees and thus a vault reference has to be set.

I agree, here we can get rid of the 0x0 check. Also in the `onlyVault` modifier, as well.

spot-contracts/contracts/RolloverVault.sol
aalavandhan: Since we tranche the entire underlying balance using the deposit bond, we usually "over-tranche" and have excess As and Zs. 

This is usually `recovered` when the next `recoverAndRedeploy` is executed but folks who redeem will get some extra A/Zs which they have to meld manually.

aalavandhan: This does increase the gas cost of weekly ops though. Can go either way with this. Open to removing, since we have the option to `recoverAndRedeem` ..

spot-contracts/contracts/PerpetualTranche.sol
brandoniles: Agree, that's probably right. Now that the vault is the only way to roll over, pause is really for "sky is falling" scenarios so we should be mindful of that.

spot-contracts/contracts/RolloverVault.sol
brandoniles: Ah got it. Let's keep it for now, but could see removing it depending on how the flows in v2 change wrt recovers.

spot-contracts/contracts/FeePolicy.sol
aalavandhan: Yea we can assume tvl*tr will be safe of overflow (tr <= 1000)

Updated to
```
uint256 juniorTR = TRANCHE_RATIO_GRANULARITY - seniorTR;
return (vaultTVL * seniorTR).mulDiv(ONE, (perpTVL * juniorTR)).mulDiv(ONE, targetSubscriptionRatio);
```

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -739,80 +763,95 @@ contract RolloverVault is
...
         }
-        _syncAsset(underlying);
+        _syncAsset(underlying_);
     }
 
     /// @dev Rolls over freshly tranched tokens from the given bond for older tranches (close to maturity) from perp.
     ///      And performs some book-keeping to keep track of the vault's assets.
     /// @return Flag indicating if any tokens were rolled over.
-    function _rollover() private returns (bool) {
+    function _rollover(IPerpetualTranche perp_, IERC20Upgradeable underlying_) private returns (bool, ITranche) {
```
brandoniles, line 839: Update @param and @return NatSpecs?

brandoniles (commented):
For #5, 294b89e28827f924fe52cc92dacb68ec453d04f3

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -281,7 +291,7 @@ contract RolloverVault is
     }
 
     /// @inheritdoc IVault
-    function recover() public override nonReentrant whenNotPaused {
+    function recover() public override whenNotPaused {
```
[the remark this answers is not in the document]
aalavandhan: I think it was because we use `recover()` in one of the swap operations. I dont think we don in the final version so it can be `nonReentrant`

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -80,6 +77,11 @@ contract RolloverVault is
     /// @dev The maximum number of deployed assets that can be held in this vault at any given time.
     uint256 public constant MAX_DEPLOYED_COUNT = 47;
 
+    /// @dev The enforced minimum number of perp or underlying tokens (in floating point units)
+    ///      that are required to be sent in during a swap.
+    ///      MIN_SWAP_UNITS = 100, means at least 100.0 perp or underlying tokens need to be swapped in.
+    uint256 public constant MIN_SWAP_UNITS = 100;
```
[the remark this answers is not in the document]
aalavandhan: Right. We don't someone to grief by swapping a really small amount of perps and the vault having to pay to deal with the tranches.

spot-contracts/contracts/RolloverVault.sol
aalavandhan: It's a private function so I don't think we need @param or @notice. The `@return` here is just for clarity

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -256,22 +258,30 @@ contract RolloverVault is
     ///      The vault holds `minUnderlyingBal` as underlying tokens and deploys the rest.
     ///      Reverts if no funds are rolled over or enforced deployment threshold is not reached.
     function deploy() public override nonReentrant whenNotPaused {
-        _deductProtocolFee();
+        IERC20Upgradeable underlying_ = underlying;
```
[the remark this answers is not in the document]
aalavandhan: A thousand or so gwei

brandoniles (approved):
LGTM

Do you know the code coverage after these unit tests?

aalavandhan:
Should be pretty close to 100%.. we'll know when the ci runs

PR #191, opened 2024-01-23
Balancer

aalavandhan:
Router to operate on spot and stAMPL together


spot-contracts/contracts/_utils/BondTranchesHelpers.sol
```diff
@@ -51,37 +53,23 @@ library BondTranchesHelpers {
...
-                minBalanceToTrancheRatio = d;
-            }
+        uint256[] memory trancheAmtsReq = new uint256[](2);
 
-            // if one of the balances is zero, we return
-            if (minBalanceToTrancheRatio <= 0) {
-                return (redeemableAmts);
-            }
-        }
+        // We compute the amount of juniors required using all the seniors
```
brandoniles, line 63: Does it make sense to first do a simple test, then an if-then?
```
if (srBal < jrBal * (srTR / jrTR) )) {
// seniors are the limiter, or correct ratio
...
} else {
// juniors are the limiter
...
}
```

Might depend on which limitation we expect to be most common. Your version should perform better when Jrs are less likely to be the limiter. The conditional check is cheap, but on failure it performs the calculations twice.

spot-contracts/contracts/_interfaces/IRolloverVault.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.0;
+
+import { IERC20Upgradeable } from "@openzeppelin/contracts-upgradeable/token/ERC20/IERC20Upgradeable.sol";
+import { IVault } from "./IVault.sol";
+import { IFeePolicy } from "./IFeePolicy.sol";
+import { SubscriptionParams } from "./ReturnData.sol";
+
+interface IRolloverVault is IVault {
```
brandoniles, line 9: NatSpecs for this file?

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -606,8 +582,12 @@ contract PerpetualTranche is
         external
         override
         afterStateUpdate
-        returns (IERC20Upgradeable[] memory, uint256[] memory)
+        returns (TokenAmount[] memory)
     {
+        uint256 perpSupply = totalSupply();
+        if(perpSupply == 0){
+            revert UnacceptableRedemption();
+        }
```
brandoniles, line 590: Change the below line to use `perpSupply`, instead of re-calling `totalSupply()`?

brandoniles (commented):
Looks good so far, I like the refactors 👍 

Covers everything except the Rebalancer

spot-contracts/contracts/_utils/BondTranchesHelpers.sol
aalavandhan: This function is called from the vault and juniors are more likely to be the limiter

spot-contracts/contracts/_interfaces/IFeePolicy.sol
```diff
@@ -84,4 +75,7 @@ interface IFeePolicy {
         uint256 vaultTVL,
         uint256 seniorTR
     ) external view returns (uint256);
+
+    /// @notice The target subscription ratio i.e) the normalization factor.
+    function targetSubscriptionRatio() external view returns (uint256);
```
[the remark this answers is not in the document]
aalavandhan: Right, we'll need this to calculate the magic ratio. (Or some method exposed by the fee policy)

spot-contracts/contracts/_utils/BondTranchesHelpers.sol
```diff
@@ -51,37 +53,23 @@ library BondTranchesHelpers {
...
-        for (i = 0; i < bt.tranches.length; i++) {
-            // NOTE: We round the avaiable balance down to the nearest multiple of the
-            //       tranche ratio. This ensures that `minBalanceToTrancheRatio`
-            //       can be represented without loss as a fixedPt number.
-            uint256 bal = trancheBalsAvailable[i] - (trancheBalsAvailable[i] % bt.trancheRatios[i]);
-            uint256 d = bal.mulDiv(TRANCHE_RATIO_GRANULARITY, bt.trancheRatios[i]);
-            if (d < minBalanceToTrancheRatio) {
-                minBalanceToTrancheRatio = d;
-            }
+        uint256[] memory trancheAmtsReq = new uint256[](2);
```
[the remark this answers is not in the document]
aalavandhan: hmm dont think that works

spot-contracts/contracts/_utils/BondTranchesHelpers.sol
aalavandhan: This function is called from the vault, where we can assume that most times the seniors are the limiter (ie vault will have more juniors).

nms-7 (approved):

spot-contracts/contracts/_interfaces/IFeePolicy.sol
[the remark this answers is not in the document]
brandoniles: Yeah. My next question (when I get there I guess) is--would the logic that compares against the targetSR be better placed in the fee strategy? Like, does this function just show that we're breaking some encapsulation that could exist.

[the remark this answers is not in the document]
aalavandhan: Right. This is what I ended up doing eventually

spot-contracts/contracts/Balancer.sol
```diff
@@ -209,18 +213,37 @@ contract Balancer {
...
+            // NOTE: Only the underlying portion of the redeemed list is used to mint perps.
+            // The rest of the tranches are returned to the user and thus we charge a redemption fee
+            // for the "non-underlying" portion.
+            uint256 vaultTVL = vault.getTVL();
+            uint256 redemptionFeeAmt = noteAmtBurnt
+                .mulDiv(vaultTVL - vault.getVaultAssetValue(perp.underlying()), vaultTVL, MathUpgradeable.Rounding.Up)
+                .mulDiv(feePolicy.computeVaultBurnFeePerc(), FEE_ONE_PERC, MathUpgradeable.Rounding.Up);
+
+            // We charge the vault fees by simply burning vault notes (without redemption)
+            IERC20Burnable(address(vault)).burn(redemptionFeeAmt);
```
aalavandhan, line 242: Maybe burn after redeem?

PR #192, opened 2024-02-01
Common types refactor

aalavandhan:
* Factored out common types
* Optimized redemption amount calc


spot-contracts/contracts/_interfaces/CommonTypes.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
brandoniles, line 1: This should be BUSL, right?

spot-contracts/contracts/_interfaces/IRolloverVault.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
brandoniles, line 1: BUSL?

brandoniles (commented):
Looks good, just the question about licensing

spot-contracts/contracts/_interfaces/CommonTypes.sol
aalavandhan: Right

spot-contracts/contracts/_interfaces/IRolloverVault.sol
aalavandhan: I think GPL-3 for interfaces are fine

brandoniles (approved):
LGTM

PR #195, opened 2024-02-19
Audit fixes - 2nd pass

aalavandhan:
Contains changes from comments in [this](https://docs.google.com/spreadsheets/d/1kDIujHNLutL0bXB4bjeFQK9n8RNehZs2lrNqilSzUus/edit#gid=0) sheet (colored in orange). 

Additionally
* Added max reserve length parameter to perp (similar to the vault)
* Updated perp's `getReserveTokensUpForRollover` method to remove `address(0)` and return a smaller array with just the tokens to rollover. It consumes slightly more gas, but lesser data is passed downstream. 
* Enforcing vault's liquidity constraints on both swap functions. (its a cheap check and there could be weird cases where the swapPerpsForUnderlying fn decreases the available vault liquidity when it does not have the matching juniors) 
* Fixed bug in the router contract, which first has to issue the deposit bond before querying it.

### DIFF
--- a/spot-contracts/contracts/FeePolicy.sol
+++ b/spot-contracts/contracts/FeePolicy.sol
@@ -3,23 +3,12 @@ pragma solidity ^0.8.20;
 
 import { IFeePolicy } from "./_interfaces/IFeePolicy.sol";
 import { SubscriptionParams } from "./_interfaces/CommonTypes.sol";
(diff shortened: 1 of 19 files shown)


nms-7 (approved):

brandoniles (approved):
LGTM

PR #199, opened 2024-03-07
Pre release fixes

aalavandhan:
* Moving back to fixed-percentage fees instead of stepfn for mint/redeem operations
* Removed vault's deployment fee
* PeckShield audit fixes
* OZ bot issue fix wrt msg.sender usage
* Minor cleanup


spot-contracts/contracts/RouterV2.sol
```diff
@@ -57,6 +58,10 @@ contract RouterV2 {
             perp.updateState();
         }
 
+        if (bond != perp.getDepositBond()) {
```
brandoniles, line 61: If we're doing a read anyway, seems there's no need to pass it in as a parameter.

WDYT about removing the `bond` parameter, then setting the function var to `perp.getDepositBond()`? Then we wouldn't need the revert condition either.

aalavandhan: The reason we pass in the parameter is to ensure the user know which bond he is depositing into (since the deposit bond can diverge between the time the user reads and the user executes).

This check here is redundant as perp would revert if you try to mint using a non-deposit tranche.

nms-7: It is an expensive read (updates state) to have since redundant

nms-7 (approved):

brandoniles (approved):
LGTM

spot-contracts/contracts/RouterV2.sol
brandoniles: Sure, we can keep as-is

PR #206, opened 2024-04-28
Removed fixed per-tranche mint cap 

aalavandhan:
* Updated deposit tranche cap to be set based on the ratio of the deposit tranche value vs the remaining tranches
* Exposing system DR on the perp and vault contracts
* Added non-reentrant to external query methods on perp


spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -887,17 +897,43 @@ contract PerpetualTranche is
...
+        uint256 depositTrancheValue = 0;
+        uint8 reserveCount = uint8(_reserves.length());
+        for (uint8 i = 1; i < reserveCount; ++i) {
+            ITranche tranche = ITranche(address(_reserveAt(i)));
+            uint256 trancheValue = _computeReserveTrancheValue(
+                tranche,
+                IBondController(tranche.bond()),
+                underlying_,
+                tranche.balanceOf(address(this)),
+                true
```
brandoniles, line 926: I view `roundUp` behavior to be the exceptional case used in particular situations, like incoming asset values to ensure positive dust.

Since we're just deciding "go/no-go" in this function and dust isn't really an issue, I'd default to the more normal math mode of truncated division.

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -887,17 +897,43 @@ contract PerpetualTranche is
...
+        // they will be eventually rolled over into the deposit tranche.
+        // If the reserve has any mature underlying tokens,
+        // we just exempt it from our calculation.
+        IERC20Upgradeable underlying_ = _reserveAt(0);
+        uint256 totalTrancheVal = 0;
+        uint256 depositTrancheValue = 0;
+        uint8 reserveCount = uint8(_reserves.length());
+        for (uint8 i = 1; i < reserveCount; ++i) {
+            ITranche tranche = ITranche(address(_reserveAt(i)));
+            uint256 trancheValue = _computeReserveTrancheValue(
```
brandoniles, line 921: I was trying to think whether we should be using tranche values or tranche balances...

I think I agree with using values--it's the more conservative one, though, since it would cut off minting earlier. It would naturally get relaxed as later tranches get closer to being overwater again.

brandoniles (commented):
Just the minor comment on the math rounding. I'm good with this accounting approach 👍

nms-7 (approved):

brandoniles (approved):
LGTM

PR #212, opened 2024-06-10
Peck audit changes

aalavandhan:


spot-vaults/contracts/BillBroker.sol
```diff
@@ -77,6 +77,7 @@ contract BillBroker is
     uint256 public constant DECIMALS = 18;
     uint256 public constant ONE = (10 ** DECIMALS);
     uint256 private constant INITIAL_RATE = 1000000;
+    uint256 public constant MINIMUM_LIQUIDITY = 10 ** 22;
```
aalavandhan, line 80: Based on our implementation on first deposit, 1e24 LP tokens  are basically 1$ worth.

Thus 1e22 = 0.01$  which is the amount we lock into the contract forever.

spot-vaults/contracts/BillBroker.sol
```diff
@@ -274,7 +279,12 @@ contract BillBroker is
         usd.safeTransferFrom(_msgSender(), address(this), usdAmtIn);
         perp.safeTransferFrom(_msgSender(), address(this), perpAmtIn);
 
-        // mint LP tokens
+        // Permanently lock the MINIMUM_LIQUIDITY tokens on first mint
+        if (isFirstMint) {
+            _mint(address(this), MINIMUM_LIQUIDITY);
```
aalavandhan, line 284: ERC20 doesn't allow to mint to address(0) anymore

spot-vaults/contracts/BillBroker.sol
```diff
@@ -274,7 +279,12 @@ contract BillBroker is
...
+        }
+
+        // mint LP tokens to the user
```
brandoniles, line 287: I think we need to subtract MINIMUM_LIQUIDITY from the amount minted to user, if it's the first mint.

(edit: answered below)

spot-vaults/contracts/BillBroker.sol
```diff
@@ -446,24 +456,31 @@ contract BillBroker is
...
         // The onus is on the depositor to ensure that the value of USD tokens and
         // perp tokens on first deposit are equivalent.
-        if (totalSupply_ <= 0) {
+        if (isFirstMint) {
             usdAmtIn = usdAmtMax;
             perpAmtIn = perpAmtMax;
             mintAmt = (ONE.mulDiv(usdAmtIn, usdUnitAmt) +
                 ONE.mulDiv(perpAmtIn, perpUnitAmt));
-            mintAmt = mintAmt * INITIAL_RATE;
+            mintAmt = (mintAmt * INITIAL_RATE) - MINIMUM_LIQUIDITY;
```
brandoniles, line 483: This works, but style-wise it seems cleaner to do the subtraction from the mint amount at the same place where the extra mint amount is minted. Colocating the two parts of this logic reduces chance of these two locations getting out of sync with each other.

When I was reading the mint function up above, it seemed like a bug because nothing was subtracted. So I think it's also less surprising to the next code reader.

aalavandhan: Moved it to the top

brandoniles (approved):
LGTM

nms-7 (approved):

PR #216, opened 2024-07-12
fix: BondIssuer.sol

nnsW3:
hello
the words were in the incorrect order, so I have revised it to make sense

Br, Elias.


aalavandhan (approved):

aalavandhan:
Thanks! LGTM

PR #217, opened 2024-07-26
Charm Vault manager update

aalavandhan:
* Fixed couple of minor bugs in the wampl-weth vault manager implementation. 1) Force rebalancing when limit ranges change. 2) Handling invalid oracle data, winding down to inf range. 3) Removing limit ranges completely instead of narrow/wide
* Added a new auto manager for the SPOT-USDC charm vault. (mostly similar to wampl-weth, but adjusts liquidity based on SPOT's deviation from FMV)

### DIFF
--- a/spot-vaults/contracts/UsdcSpotManager.sol
+++ b/spot-vaults/contracts/UsdcSpotManager.sol
@@ -125,7 +125,7 @@ contract UsdcSpotManager {
 
     /// @notice Executes vault rebalance.
     function rebalance() public {
-        (uint256 deviation, bool deviaitonValid) = computeDeviationFactor();
+        (uint256 deviation, bool deviationValid) = computeDeviationFactor();
 
         // We rebalance if the deviation factor has crossed ONE (in either direction).
         bool forceLiquidityUpdate = ((deviation <= ONE && prevDeviation > ONE) ||
@@ -141,7 +141,7 @@ contract UsdcSpotManager {
         // the vault sells SPOT and deviation is above ONE, or when
         // the vault buys SPOT and deviation is below ONE
         bool extraSpot = isOverweightSpot();
-        bool activeLimitRange = deviaitonValid &&
+        bool activeLimitRange = deviationValid &&
             ((deviation >= ONE && extraSpot) || (deviation <= ONE && !extraSpot));
 
         // Trim positions after rebalance.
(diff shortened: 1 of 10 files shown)


spot-vaults/contracts/WethWamplManager.sol
```diff
@@ -241,44 +245,52 @@ contract WethWamplManager {
 
     /// @notice Executes vault rebalance.
     function rebalance() public {
-        uint256 deviation = computeDeviationFactor();
-        uint256 activeLiqPerc = computeActiveLiqPerc(deviation);
+        (uint256 deviation, bool deviaitonValid) = computeDeviationFactor();
+        uint256 activeLiqPerc = deviaitonValid
```
brandoniles, line 249: Maybe move this line below to the "execute rebalance" sections, since it's not used in the "update limit range" section

spot-vaults/contracts/UsdcSpotManager.sol
```diff
@@ -0,0 +1,228 @@
...
+
+    /// @notice Constructor initializes the contract with provided addresses.
+    /// @param vault_ Address of the AlphaProVault contract.
+    /// @param spotAppraiser_ Address of the spot appraiser.
+    constructor(IAlphaProVault vault_, IBillBrokerPricingStrategy spotAppraiser_) {
+        owner = msg.sender;
+
+        VAULT = vault_;
+        POOL = vault_.pool();
+        USDC = vault_.token0();
```
brandoniles, line 93: Probably better style to hardcode the expected token addresses, then do a `require` check on the vault properties on init.

Nothing inherently incorrect with this approach, provided we do a manual spot check on deployment.

spot-vaults/contracts/UsdcSpotManager.sol
```diff
@@ -0,0 +1,228 @@
...
+        require(success, "Vault call failed");
+        return r;
+    }
+
+    //--------------------------------------------------------------------------
+    // External write methods
+
+    /// @notice Executes vault rebalance.
+    function rebalance() public {
+        (uint256 deviation, bool deviaitonValid) = computeDeviationFactor();
```
brandoniles, line 154:
```suggestion
        (uint256 deviation, bool deviationValid) = computeDeviationFactor();
```
(and other uses)

spot-vaults/contracts/UsdcSpotManager.sol
```diff
@@ -0,0 +1,228 @@
...
+        // The limit range has more SPOT than USDC if `_marketPrice >= _limitPrice`,
+        // so the vault looks to sell SPOT.
+        // NOTE: This assumes that in the underlying univ3 pool
+        // token0 is USDC and token1 is SPOT.
+        bool isSpotAsk = (_marketPrice >= _limitPrice);
+        return ((deviation >= ONE && isSpotAsk) || (deviation <= ONE && !isSpotAsk));
+    }
+
+    /// @return The computed SPOT price in USD from the underlying univ3 pool.
+    function getSpotUSDPrice() public view returns (uint256) {
```
brandoniles, line 205: This is perhaps the first time any code we've written relies on realtime market prices from dexes.

I think the security risk is minimal, since it would require an attacker to manipulate the spot market to profit from forcing limit range changes, right? Not sure I see any benefits, but calling out just in case.

aalavandhan: We will be using a 1hr TWAP here. The wampl vault has been configured to use a 1hr twap, the spot vault needs to be updated.

spot-vaults/contracts/WethWamplManager.sol
aalavandhan: Restructured with the newer version

brandoniles (commented):
LGTM, just the comment update and looks good

brandoniles (approved):

PR #218, opened 2024-08-01
Single sided deposit for Billy

aalavandhan:
1) Added USD,Perp single sided deposit when pool is unbalanced
2) Added events for subgraph indexing

### DIFF
--- a/spot-vaults/contracts/BillBroker.sol
+++ b/spot-vaults/contracts/BillBroker.sol
@@ -294,19 +294,33 @@ contract BillBroker is
 
     /// @notice Single sided usd token deposit and mint LP tokens.
     /// @param usdAmtIn The amount of usd tokens to be deposited.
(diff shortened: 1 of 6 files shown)


spot-vaults/contracts/BillBroker.sol
```diff
@@ -508,6 +577,84 @@ contract BillBroker is
...
+                usdBalance: s.usdBalance,
+                perpBalance: s.perpBalance + perpAmtIn,
+                usdPrice: s.usdPrice,
+                perpPrice: s.perpPrice
+            })
+        );
+
+        // We only allow minting LP-tokens with Perps,
+        // when asset ratio is greater than ONE; i.e) pool is overweight USD.
+        if (assetRatioPre <= ONE || assetRatioPost <= ONE) {
```
brandoniles, line 644:
```suggestion
        if (assetRatioPre <= ONE || assetRatioPost < ONE) {
```

PR #220, opened 2024-08-09
Spot pricer for automanager

aalavandhan:
Added a new SPOT pricing strategy for the charm manager. 

The bill broker uses the spot appraiser which has stricter rules around DR and tranche CDRs. However, the AMM pool can use a much simpler pricing strategy.

### DIFF
--- a/spot-vaults/contracts/_strategies/SpotCDRPricer.sol
+++ b/spot-vaults/contracts/_strategies/SpotCDRPricer.sol
@@ -1,42 +1,30 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.24;
 
-import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
 import { Math } from "@openzeppelin/contracts/utils/math/Math.sol";
 
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import { ITranche } from "@ampleforthorg/spot-contracts/contracts/_interfaces/buttonwood/ITranche.sol";
-import { IBondController } from "@ampleforthorg/spot-contracts/contracts/_interfaces/buttonwood/IBondController.sol";
 import { IPerpetualTranche } from "@ampleforthorg/spot-contracts/contracts/_interfaces/IPerpetualTranche.sol";
 import { IChainlinkOracle } from "../_interfaces/external/IChainlinkOracle.sol";
 import { IAmpleforthOracle } from "../_interfaces/external/IAmpleforthOracle.sol";
 import { ISpotPricingStrategy } from "../_interfaces/ISpotPricingStrategy.sol";
-import { InvalidSeniorCDRBound } from "../_interfaces/BillBrokerErrors.sol";
 
 /**
- * @title SpotAppraiser
+ * @title SpotCDRPricer
  *
- * @notice Pricing strategy adapter for a BillBroker vault which accepts
- *         SPOT (as the perp token) and dollar tokens like USDC.
+ * @notice Pricing strategy adapter for SPOT.
  *
- *         AMPL is the underlying token for SPOT.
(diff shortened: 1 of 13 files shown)


nms-7 (approved):

spot-vaults/contracts/_strategies/SpotCDRPricer.sol
```diff
@@ -0,0 +1,125 @@
...
+    // External methods
+
+    /// @return p The price of the usd token in dollars.
+    /// @return v True if the price is valid and can be used by downstream consumers.
+    function usdPrice() external view override returns (uint256, bool) {
+        (uint256 p, bool v) = _getCLOracleData(USD_ORACLE, USD_ORACLE_DECIMALS);
+        // If the market price of the USD coin fallen too much below 1$,
+        // it's an indication of some systemic issue with the USD token
+        // and thus its price should be considered unreliable.
+        return (ONE, (v && p > USD_LOWER_BOUND));
```
brandoniles, line 91: WDYT about checking deviation USDC from $1.00, which also covers the case where the dollar token is >> $1? That's also a sign of market trouble, that's missed by just checking the downside.

aalavandhan: I don't think that's specifically a risk when it comes to USDC. If usdc is trading above one, folks will just mint more? Custody/Bank failure risk is only on the downside.

brandoniles: I'm more coming from the perspective of-- IF the price of USDC is high, then something is going wrong.

Perhaps minting is halted by Circle corp (there are humans in the loop there), a liquidity pool address gets blacklisted through their AML system, there's an chainlink oracle attack/failure. Maybe other things as well I can't think of as well.

aalavandhan: Sure. Worth noting, that the current spot appraiser which is deployed and used by Billy doesn't have this check.

spot-vaults/contracts/_strategies/SpotCDRPricer.sol
```diff
@@ -36,6 +36,7 @@ contract SpotCDRPricer is ISpotPricingStrategy {
     uint256 private constant ONE = (10 ** DECIMALS);
     uint256 public constant CL_ORACLE_DECIMALS = 8;
     uint256 public constant CL_ORACLE_STALENESS_THRESHOLD_SEC = 3600 * 48; // 2 days
+    uint256 public constant USD_LOWER_UPPER = (101 * ONE) / 100; // 1.01$
```
brandoniles, line 39:
```suggestion
    uint256 public constant USD_UPPER_BOUND = (101 * ONE) / 100; // 1.01$
```

spot-vaults/contracts/_strategies/SpotCDRPricer.sol
```diff
@@ -85,10 +86,10 @@ contract SpotCDRPricer is ISpotPricingStrategy {
     /// @return v True if the price is valid and can be used by downstream consumers.
     function usdPrice() external view override returns (uint256, bool) {
         (uint256 p, bool v) = _getCLOracleData(USD_ORACLE, USD_ORACLE_DECIMALS);
-        // If the market price of the USD coin fallen too much below 1$,
+        // If the market price of the USD coin deviated too much from 1$,
         // it's an indication of some systemic issue with the USD token
         // and thus its price should be considered unreliable.
-        return (ONE, (v && p > USD_LOWER_BOUND));
+        return (ONE, (v && p < USD_LOWER_UPPER && p > USD_LOWER_BOUND));
```
brandoniles, line 92:
```suggestion
        return (ONE, (v && p < USD_UPPER_BOUND && p > USD_LOWER_BOUND));
```

brandoniles (approved):
LGTM, with the slight name change

PR #221, opened 2024-08-29
Swap limit bug fix

aalavandhan:
Skipping liquidity check when swapping from perps to underlying.

Swapping from perps to underlying should almost always increase the vault's liquidity, so we only check if underlying balance doesn't decrease.

### DIFF
--- a/spot-contracts/contracts/RolloverVault.sol
+++ b/spot-contracts/contracts/RolloverVault.sol
@@ -440,8 +440,9 @@ contract RolloverVault is
     function swapUnderlyingForPerps(uint256 underlyingAmtIn) external nonReentrant whenNotPaused returns (uint256) {
         // Calculates the fee adjusted perp amount to transfer to the user.
         // NOTE: This operation should precede any token transfers.
-        IERC20Upgradeable underlying_ = underlying;
         IPerpetualTranche perp_ = perp;
+        IERC20Upgradeable underlying_ = underlying;
+        uint256 underlyingBalPre = underlying_.balanceOf(address(this));
         (uint256 perpAmtOut, uint256 perpFeeAmtToBurn, SubscriptionParams memory s) = computeUnderlyingToPerpSwapAmt(
             underlyingAmtIn
         );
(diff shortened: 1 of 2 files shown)


nms-7 (approved):

nms-7 (approved):

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -507,8 +509,11 @@ contract RolloverVault is
         // transfer underlying out
         underlying_.safeTransfer(msg.sender, underlyingAmtOut);
 
-        // Revert if vault liquidity is too low.
-        _enforceUnderlyingBalAfterSwap(underlying_, s.vaultTVL);
+        // Revert if swap reduces vault's available liquidity.
+        uint256 underlyingBalPost = underlying_.balanceOf(address(this));
+        if (underlyingBalPost <= underlyingBalPre) {
```
brandoniles, line 514:
```suggestion
        if (underlyingBalPost < underlyingBalPre) {
```
Should be ok if it remains unchanged?

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -469,7 +469,12 @@ contract RolloverVault is
         // The vault continues to hold the perp dust until the subsequent `swapPerpsForUnderlying` or manual `recover(perp)`.
 
         // Revert if vault liquidity is too low.
-        _enforceUnderlyingBalAfterSwap(underlying_, s.vaultTVL);
+        //  - Absolute balance is strictly greater than `minUnderlyingBal`.
+        //  - Ratio of the balance to the vault's TVL is strictly greater than `minUnderlyingPerc`.
+        uint256 underlyingBal = underlying_.balanceOf(address(this));
```
brandoniles, line 474: I guess there's a weird corner case where underlying balance could go up (or maybe just remain constant?) after an underlying->perp swap.
e.g. If the vault was sitting on a bunch of As, and the new Zs free those up.
In that case I think we'd want to allow that to go through.

Perhaps the most generalized logic would be:

```
If underlying bal increases or stays the same, continue
else, execute the check logic you have here.
```
That situation would be exceptionally rare, though. Worth it? It would be one extra var and a conditional, I think.

aalavandhan: cool. So basically if the underlying balances goes down, we then ensure that it doesn't go down too much. If it stays the same or increases its fine..

brandoniles (approved):
LGTM

PR #222, opened 2024-09-08
Updated liquidity check condition based on system subscription ratio

aalavandhan:
1) Vault's min liquidity constraints are enforced universally before rollover and after swaps
2) The limiting percentage parameter is a percentage of vault's neutral TVL


nms-7 (approved):

brandoniles (approved):

brandoniles:
Approving, but assuming we'll recommend setting this to zero

aalavandhan:
Yes. We'll rely on existing dr checks to limit swaps. We can view this an emergency leaver, to limit rollovers.

PR #229, opened 2024-10-17
Updated CDR pricer

aalavandhan:
Newly deployed CDR pricer which uses policy to the cpi adjusted ampl target rate


brandoniles (approved):
LGTM

PR #231, opened 2024-10-28
Scalable usdc-spot univ3 strategy

aalavandhan:
An new alpha vault strategy for spot-usdc which is scalable and can provide deep concentrated liquidity around SPOT's fmv.

### DIFF
--- a/spot-vaults/contracts/charm/UsdcSpotManager.sol
+++ b/spot-vaults/contracts/charm/UsdcSpotManager.sol
@@ -1,8 +1,11 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.24;
 
+import { Math } from "@openzeppelin/contracts/utils/math/Math.sol";
+import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
 import { AlphaVaultHelpers } from "../_utils/AlphaVaultHelpers.sol";
+import { Range } from "../_interfaces/types/CommonTypes.sol";
 
 import { IMetaOracle } from "../_interfaces/IMetaOracle.sol";
 import { IAlphaProVault } from "../_interfaces/external/IAlphaProVault.sol";
@@ -10,10 +13,20 @@ import { IUniswapV3Pool } from "@uniswap/v3-core/contracts/interfaces/IUniswapV3
 
 /// @title UsdcSpotManager
 /// @notice This contract is a programmatic manager for the USDC-SPOT Charm AlphaProVault.
+///
+/// @dev The vault's active zone is defined as lower and upper percentages of FMV.
+///      For example, if the active zone is [0.95, 1.05]x and SPOT's FMV price is $1.35.
+///      When the market price of SPOT is between [$1.28, $1.41] we consider price to be in the active zone.
+///
+///      When in the active zone, the vault provides concentrated liquidity around the market price.
+///      When price is outside the active zone, the vault reverts to a full range position.
+///
+///
 contract UsdcSpotManager is Ownable {
     //-------------------------------------------------------------------------
     // Libraries
     using AlphaVaultHelpers for IAlphaProVault;
+    using Math for uint256;
 
     //-------------------------------------------------------------------------
     // Constants & Immutables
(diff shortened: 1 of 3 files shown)


spot-vaults/contracts/charm/UsdcSpotManager.sol
```diff
@@ -37,6 +48,17 @@ contract UsdcSpotManager is Ownable {
     /// @notice The recorded deviation factor at the time of the last successful rebalance operation.
     uint256 public prevDeviation;
 
+    /// @notice The lower and upper deviation factor within which
+    ///         SPOT's price is considered to be in the active zone.
```
brandoniles, line 52: Looking at the logic further down, this comment isn't exactly accurate.

This zone determines the max deviation where concentrated liquidity bands can be placed. So, the deviation factor range is smaller than this range. This distinction will be important for setting the configuration correctly.

spot-vaults/contracts/charm/UsdcSpotManager.sol
```diff
@@ -144,8 +168,55 @@ contract UsdcSpotManager is Ownable {
...
+        );
+        uint256 limitBandDeviationWidth = isOverweightSpot()
+            ? activeZoneDeviation.upper - deviation
+            : deviation - activeZoneDeviation.lower;
+        VAULT.setFullRangeWeight(SafeCast.toUint24(fullRangeWeight));
+
+        // IMPORTANT: The deviation percentage, to ticks conversion isn't precise.
+        // In the worst case, there might be liquidity up to 2% outside the active zone definition.
+        VAULT.setBaseThreshold(deviationToTicks(concBandDeviationWidth));
+        VAULT.setLimitThreshold(deviationToTicks(limitBandDeviationWidth));
```
aalavandhan, line 200: We can consider removing the limit ranges completely when it's buying above fmv and selling below. It does slightly reduce the opportunity of getting back to 50/50 in the long run, however it also eliminates IL. 
But considering the active zone is quite tight to begin with maybe it doesn't matter.

brandoniles: If we can, we should have the logic make sense independent of configuration. At this stage, it's best if we don't make any assumptions about how tight the active zone is

aalavandhan: Right, makes sense to remove it then.

aalavandhan: Okay, will go back to adding limit ranges only based on fmv

spot-vaults/contracts/charm/UsdcSpotManager.sol
```diff
@@ -59,6 +67,9 @@ contract UsdcSpotManager is Ownable {
     /// @notice The maximum USDC balance of the vault's full range position.
     uint256 public fullRangeMaxUsdcBal;
 
+    /// @notice The maximum percentage of vault's balanced assets in the full range position.
+    uint256 public fullRangeMaxPerc;
```
aalavandhan, line 71: Both absolute and percentage limits will help minimize intervention

spot-vaults/contracts/charm/UsdcSpotManager.sol
aalavandhan: Okay, rethinking this framing re IL. I think the original implementation makes the most sense. We keep concentrated/limit liquidity when in the active zone and revert to bounded infinite band when outside.

spot-vaults/contracts/charm/UsdcSpotManager.sol
aalavandhan: updated the comment

spot-vaults/contracts/charm/UsdcSpotManager.sol
```diff
@@ -87,23 +113,50 @@ contract UsdcSpotManager is Ownable {
...
+    }
+
     //--------------------------------------------------------------------------
     // External write methods
 
     /// @notice Executes vault rebalance.
     function rebalance() public {
         (uint256 deviation, bool deviationValid) = oracle.spotPriceDeviation();
+        bool withinActiveZone = (deviationValid && activeZone(deviation));
+        bool withinActiveZoneBefore = activeZone(prevDeviation);
```
brandoniles, line 145: Consider caching the boolean `withinActiveZoneBefore`, rather than recomputing it from the cached prevDeviation each time.

aalavandhan: done

brandoniles (approved):
LGTM

PR #232, opened 2024-11-05
Fee curve update

aalavandhan:
Updated enrichment/debasement fee curve. 

**Motivation:**

The current implementation caps the enrichment rate when dr is high. The side-effect of this is that the vault's effective funding rate (rate vault holders pay to get magnified exposure) reduces. 

When dr = 2, for example the perpTVL:vaultTVL ratio is 1:6 and at dr = 3 its 1:9. Say there's a perp tvl of 1m$. Perp's enrichment would be ~200k at dr = 2 and at dr = 3. However, the vault's effective rate would be 200k/6m ~3.3% when dr =2 and 200k/9m ~2.2% when dr = 3 and so on. So it actually continues to go down. (As shown in the following figure)

download

Alternatively as dr continues to increase we want the vault's effective funding rate to increase or at least stay constant. This would mean that perp's enrichment rate is unbounded on the upside.
(As shown below)
download

### DIFF
--- a/spot-contracts/contracts/FeePolicy.sol
+++ b/spot-contracts/contracts/FeePolicy.sol
@@ -3,13 +3,11 @@ pragma solidity ^0.8.20;
 
 import { IFeePolicy } from "./_interfaces/IFeePolicy.sol";
 import { SubscriptionParams } from "./_interfaces/CommonTypes.sol";
-import { InvalidPerc, InvalidTargetSRBounds, InvalidDRBounds, InvalidSigmoidAsymptotes } from "./_interfaces/ProtocolErrors.sol";
+import { InvalidPerc, InvalidTargetSRBounds, InvalidDRBounds } from "./_interfaces/ProtocolErrors.sol";
 
 import { MathUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/math/MathUpgradeable.sol";
 import { SafeCastUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/math/SafeCastUpgradeable.sol";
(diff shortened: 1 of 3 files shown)


spot-contracts/contracts/FeePolicy.sol
```diff
@@ -100,17 +94,19 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
-    struct RolloverFeeSigmoidParams {
-        /// @notice Lower asymptote
-        int256 lower;
-        /// @notice Upper asymptote
-        int256 upper;
-        /// @notice sigmoid slope
-        int256 growth;
+    struct RolloverFeeParams {
+        /// @notice The maximum debasement rate for perp,
+        ///         i.e) the maximum rate perp pays the vault for rollovers.
```
aalavandhan, line 99: Changing types here, but should be safe

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -274,14 +263,16 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
     function computePerpRolloverFeePerc(uint256 dr) external view override returns (int256) {
-        return
-            Sigmoid.compute(
-                dr.toInt256(),
-                perpRolloverFee.lower,
-                perpRolloverFee.upper,
-                perpRolloverFee.growth,
-                ONE.toInt256()
+        if (dr <= ONE) {
+            uint256 perpRate = MathUpgradeable.min(
```
brandoniles, line 267: Let's call this something like `negPerpRate` (most accurate?) so it's more clear what this intermediate value is.

Then, maybe take the max of `-1* negPerpRate` and the debasement limit? Seems more in the spirit of the computation.

brandoniles (approved):
LGTM

PR #237, opened 2025-01-07
Bill broker fee curve change

aalavandhan:
Previously the bill broker swap fee curve was piecewise function with 2 lines. This change makes it a piecewise with 3 lines as shown in the image and now supports negative fees. 

We stored fee percentages as a unsigned fixed-point number with 18 decimals (1e18 = 100% or 1.0) which doesn't support negative fee values. Now we simply covert the existing percentages into a factor by adding 1e18. Under the new scheme:
0% => 1.0 or 1e18
100% => 2.0 or 2e18
-20% => 0.8 or 8e17

image1
image

### DIFF
--- a/spot-vaults/contracts/BillBroker.sol
+++ b/spot-vaults/contracts/BillBroker.sol
@@ -662,9 +662,13 @@ contract BillBroker is
         // Compute mint amount.
         mintAmt = valueIn.mulDiv(totalSupply_, totalReserveVal);
 
-        // Apply a combination of swap and mint fees.
+        // A single sided deposit is a combination of swap and mint.
+        // We first calculate the amount of usd swapped into perps and
+        // apply the swap fee only for that portion.
+        // The mint fees are waived, because single sided deposits
+        // push the pool back into balance.
         uint256 percOfAmtInSwapped = ONE.mulDiv(
-            usdAmtIn - mintAmt.mulDiv(s.usdBalance, totalSupply_),
+            usdAmtIn - (s.usdBalance + usdAmtIn).mulDiv(mintAmt, totalSupply_),
             usdAmtIn
         );
         uint256 feeFactor = computeUSDToPerpSwapFeeFactor(
@@ -674,7 +678,6 @@ contract BillBroker is
         mintAmt =
             mintAmt.mulDiv(percOfAmtInSwapped, ONE).mulDiv(TWO - feeFactor, ONE) +
             mintAmt.mulDiv(ONE - percOfAmtInSwapped, ONE);
-        mintAmt = mintAmt.mulDiv(ONE - fees.mintFeePerc, ONE);
     }
 
     /// @notice Computes the amount of LP tokens minted,
@@ -704,9 +707,13 @@ contract BillBroker is
             ? valueIn.mulDiv(totalSupply_, totalReserveVal)
             : 0;
 
-        // Apply a combination of swap and mint fees.
+        // A single sided deposit is a combination of swap and mint.
+        // We first calculate the amount of perps swapped into usd and
+        // apply the swap fee only for that portion.
+        // The mint fees are waived, because single sided deposits
+        // push the pool back into balance.
         uint256 percOfAmtInSwapped = ONE.mulDiv(
-            perpAmtIn - mintAmt.mulDiv(s.perpBalance, totalSupply_),
+            perpAmtIn - (s.perpBalance + perpAmtIn).mulDiv(mintAmt, totalSupply_),
             perpAmtIn
         );
         uint256 feeFactor = computePerpToUSDSwapFeeFactor(
@@ -716,7 +723,6 @@ contract BillBroker is
         mintAmt =
             mintAmt.mulDiv(percOfAmtInSwapped, ONE).mulDiv(TWO - feeFactor, ONE) +
             mintAmt.mulDiv(ONE - percOfAmtInSwapped, ONE);
-        mintAmt = mintAmt.mulDiv(ONE - fees.mintFeePerc, ONE);
     }
 
     /// @notice Computes the amount of usd and perp tokens redeemed,
@@ -829,23 +835,29 @@ contract BillBroker is
             return MAX_FEE_FACTOR;
         }
 
-        Range memory f1 = fees.perpToUSDSwapFeeFactors;
-        Range memory f2 = fees.usdToPerpSwapFeeFactors;
+        Range memory feeP2U = fees.perpToUSDSwapFeeFactors;
+        Range memory feeU2P = fees.usdToPerpSwapFeeFactors;
+        uint256 feeU2PDelta = feeU2P.upper - feeU2P.lower;
         return
             LineHelpers
                 .computePiecewiseAvgY(
                     Line({
                         x1: arHardBound.lower,
-                        y1: f1.upper,
+                        y1: feeP2U.upper,
                         x2: arSoftBound.lower,
-                        y2: f1.lower
+                        y2: feeP2U.lower
                     }),
-                    Line({ x1: 0, y1: f1.lower, x2: ONE, y2: f1.lower }),
+                    Line({
+                        x1: arSoftBound.lower,
+                        y1: feeP2U.lower,
+                        x2: arSoftBound.upper,
+                        y2: feeP2U.lower
+                    }),
                     Line({
                         x1: arSoftBound.upper,
-                        y1: f1.lower,
+                        y1: feeP2U.lower,
                         x2: arHardBound.upper,
-                        y2: (f1.lower + f2.lower - f2.upper)
+                        y2: (feeP2U.lower - feeU2PDelta)
                     }),
                     arSoftBound,
                     Range({ lower: arPost, upper: arPre })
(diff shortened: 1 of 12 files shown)


spot-vaults/contracts/BillBroker.sol
```diff
@@ -775,122 +806,90 @@ contract BillBroker is
...
-        //    |         \     |
-        // fl |          \__________
-        //    |               |
-        //    |               |
-        //    |               |
-        //    +---------------------------> ar
-        //       arHL  arSL  1.0
-        //
-        Range memory swapFeePercs = fees.perpToUSDSwapFeePercs;
+        Range memory f1 = fees.perpToUSDSwapFeeFactors;
```
brandoniles, line 832: Can we rename f1 and f2 to have a reference to perpToUSD or USDToPerp somehow?

Can be abbreviated, like feeP2U or feeU2P

spot-vaults/contracts/_utils/LineHelpers.sol
```diff
@@ -30,19 +53,72 @@ library LineHelpers {
...
+        Range memory xRange
+    ) internal pure returns (int256) {
+        int256 xl = xRange.lower.toInt256();
+        int256 xu = xRange.upper.toInt256();
+        int256 bpl = xBreakPt.lower.toInt256();
+        int256 bpu = xBreakPt.upper.toInt256();
+
+        // Validate range inputs (custom errors omitted here).
+        if (xl > xu) revert InvalidRange();
+        if (xl <= bpl && xu > bpu) revert UnexpectedRangeDelta();
```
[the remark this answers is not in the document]
aalavandhan: Could be. I just chose to keep the implementation simple in the first iteration. We can add a case F in the next pass.

spot-vaults/contracts/BillBroker.sol
```diff
@@ -227,8 +237,10 @@ contract BillBroker is
         if (
             fees_.mintFeePerc > ONE ||
             fees_.burnFeePerc > ONE ||
-            fees_.perpToUSDSwapFeePercs.lower > fees_.perpToUSDSwapFeePercs.upper ||
-            fees_.usdToPerpSwapFeePercs.lower > fees_.usdToPerpSwapFeePercs.upper ||
+            fees_.perpToUSDSwapFeeFactors.lower > fees_.perpToUSDSwapFeeFactors.upper ||
+            fees_.perpToUSDSwapFeeFactors.lower < ONE ||
```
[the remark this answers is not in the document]
aalavandhan: Right.

spot-vaults/contracts/BillBroker.sol
```diff
@@ -633,22 +645,36 @@ contract BillBroker is
...
+        // Apply a combination of swap and mint fees.
+        uint256 percOfAmtInSwapped = ONE.mulDiv(
+            usdAmtIn - mintAmt.mulDiv(s.usdBalance, totalSupply_),
+            usdAmtIn
+        );
+        uint256 feeFactor = computeUSDToPerpSwapFeeFactor(
+            assetRatio(s),
+            assetRatio(_updatedReserveState(s, s.usdBalance + usdAmtIn, s.perpBalance))
+        );
+        mintAmt =
```
brandoniles, line 674: Let's waive the mint fee on single sided deposits. It's good to encourage depositors who balance out the pool.

spot-vaults/contracts/BillBroker.sol
```diff
@@ -633,22 +645,36 @@ contract BillBroker is
...
+        if (totalReserveVal == 0 || totalSupply_ == 0) {
+            return 0;
+        }
+
+        // Compute mint amount.
+        mintAmt = valueIn.mulDiv(totalSupply_, totalReserveVal);
+
+        // Apply a combination of swap and mint fees.
+        uint256 percOfAmtInSwapped = ONE.mulDiv(
+            usdAmtIn - mintAmt.mulDiv(s.usdBalance, totalSupply_),
```
brandoniles, line 667:
```suggestion
            usdAmtIn - mintAmt.mulDiv(s.usdBalance + usdAmtIn, totalSupply_),
```
This more correct?

spot-vaults/contracts/BillBroker.sol
```diff
@@ -775,122 +806,90 @@ contract BillBroker is
...
-            );
+            LineHelpers
+                .computePiecewiseAvgY(
+                    Line({
+                        x1: arHardBound.lower,
+                        y1: f1.upper,
+                        x2: arSoftBound.lower,
+                        y2: f1.lower
+                    }),
+                    Line({ x1: 0, y1: f1.lower, x2: ONE, y2: f1.lower }),
```
brandoniles, line 843:
```suggestion
                    Line({ x1: arSoftBound.lower, y1: f1.lower, x2: arSoftBound.upper, y2: f1.lower }),
```
This is easier to visualize I think

spot-vaults/contracts/BillBroker.sol
```diff
@@ -775,122 +806,90 @@ contract BillBroker is
...
+                    Line({
+                        x1: arSoftBound.upper,
+                        y1: f1.lower,
+                        x2: arHardBound.upper,
+                        y2: (f1.lower + f2.lower - f2.upper)
```
brandoniles, line 848:
```suggestion
                        y2: (f1.lower - f2.upper + f2.lower)
```

spot-vaults/contracts/BillBroker.sol
```diff
@@ -662,9 +662,13 @@ contract BillBroker is
...
 
-        // Apply a combination of swap and mint fees.
+        // A single sided deposit is a combination of swap and mint.
+        // We first calculate the amount of usd swapped into perps and
+        // apply the swap fee only for that portion.
+        // The mint fees are waived, because single sided deposits
+        // push the pool back into balance.
         uint256 percOfAmtInSwapped = ONE.mulDiv(
-            usdAmtIn - mintAmt.mulDiv(s.usdBalance, totalSupply_),
+            usdAmtIn - (s.usdBalance + usdAmtIn).mulDiv(mintAmt, totalSupply_),
```
brandoniles, line 671: Curious how you derived this. I get...
usdAmtIn * (1 - s.usdBalance / (s.usdBalance + s.perpBalance))

Basically applying the current asset ratio to the incoming USD amount. Maybe they're the same, but it feels more natural to use asset balances rather than LP-notes for some reason.

spot-vaults/contracts/BillBroker.sol
```diff
@@ -829,23 +835,29 @@ contract BillBroker is
...
+                        y2: feeP2U.lower
                     }),
-                    Line({ x1: 0, y1: f1.lower, x2: ONE, y2: f1.lower }),
                     Line({
                         x1: arSoftBound.upper,
-                        y1: f1.lower,
+                        y1: feeP2U.lower,
                         x2: arHardBound.upper,
-                        y2: (f1.lower + f2.lower - f2.upper)
+                        y2: (feeP2U.lower - feeU2PDelta)
```
brandoniles, line 860: Design note:
If we want these fee curves to "converge" at the hard bounds, then we could change this one line to `-feeU2P.upper` I think?
We wouldn't have to change any storage slots. I need to think more before I actually advocate for this, though...

spot-vaults/contracts/BillBroker.sol
```diff
@@ -868,23 +880,29 @@ contract BillBroker is
...
+        Range memory feeU2P = fees.usdToPerpSwapFeeFactors;
+        Range memory feeP2U = fees.perpToUSDSwapFeeFactors;
+        uint256 feeP2UDelta = feeP2U.upper - feeP2U.lower;
         return
             LineHelpers
                 .computePiecewiseAvgY(
                     Line({
                         x1: arSoftBound.lower,
-                        y1: f1.lower,
+                        y1: feeU2P.lower,
```
brandoniles, line 891: I would swap (x1,y1) with (x2,y2) here, so that we're always having the line segments go rightward on the x-axis. I think this should be mathematically equivalent.

spot-vaults/contracts/BillBroker.sol
aalavandhan: yea I'd also need to think about it more

spot-vaults/contracts/BillBroker.sol
aalavandhan: hmm. You'd have to value the incoming tokens based on both the usd and perp balance (and their current prices) of the reserve. We do that do calculate `mintAmt` which is `valueIn / tvl * totalSupply`.

Now with `mintAmt`, we know what share of the reserve assets the user has a claim on.

usdClaim = mintAmt / totalSupply * (s.usdBalance + usdAmtIn)
percUsdSwapped = (usdAmtIn - usdClaim) / usdAmtIn

brandoniles: I think you can factor out the prices, kind of like how any non-oraclized DEX pools works. Any normal double-sided deposit is an op that doesn't change the asset ratio of the system, so you can just look at token balances.

aalavandhan: I dnot think so? Prices might have diverged since the last interaction. (fmv can change for example)

brandoniles: Got it. These two lines from your comment above clicked for me:
```
usdClaim = mintAmt / totalSupply * (s.usdBalance + usdAmtIn)
percUsdSwapped = (usdAmtIn - usdClaim) / usdAmtIn
```
From there, it's easy to see it in the code. If you have a way to communicate that in the code comments that's great 👍

spot-vaults/contracts/BillBroker.sol
aalavandhan: This should nudge the pool more back into balance. There's just a question of is it giving too much away? Hard to say. If we intend to lower the "feeLower" (on both directions), from 2.5% to say 1% anyway, then the impact of this should be marginal. 

No strong feelings, but thinking of leaving it as is and revisiting later if need be. Curios to hear what @gemini thinks ..

brandoniles (commented):
Just the one comment on MIN & MAX configuration, but looks gtg otherwise 👍

spot-vaults/contracts/BillBroker.sol
```diff
@@ -237,11 +232,26 @@ contract BillBroker is
...
+            fees_.perpToUSDSwapFeeFactors.upper > (TWO - MIN_FEE_FACTOR)
+        ) {
+            revert InvalidPerc();
+        }
+
+        if (
             fees_.usdToPerpSwapFeeFactors.lower > fees_.usdToPerpSwapFeeFactors.upper ||
             fees_.usdToPerpSwapFeeFactors.lower < ONE ||
-            fees_.protocolSwapSharePerc > ONE
+            fees_.usdToPerpSwapFeeFactors.upper > (TWO - MIN_FEE_FACTOR)
```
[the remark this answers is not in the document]
aalavandhan: done

brandoniles (approved):

PR #240, opened 2025-02-28
Rollover fee curve update

aalavandhan:
The rollover fee computation adjusts the fees based on the vault's subscription ratio.
For example, if the vault's sr = 0.5, it means that at most half of perp can be rolled over. Thus the rollover fees are doubled.

### DIFF
--- a/spot-contracts/contracts/FeePolicy.sol
+++ b/spot-contracts/contracts/FeePolicy.sol
@@ -96,21 +96,21 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
 
     /// @dev NOTE: We updated the type of the parameters from int256 to uint256, which is an upgrade safe operation.
     struct RolloverFeeParams {
-        /// @notice The maximum debasement rate for perp,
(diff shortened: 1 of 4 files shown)


spot-contracts/contracts/FeePolicy.sol
```diff
@@ -96,21 +96,21 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
 
     /// @dev NOTE: We updated the type of the parameters from int256 to uint256, which is an upgrade safe operation.
     struct RolloverFeeParams {
-        /// @notice The maximum debasement rate for perp,
-        ///         i.e) the maximum rate perp pays the vault for rollovers.
+        /// @notice The minimum rollover fee percentage enforced by the contract.
+        /// @dev This is represented as signed fixed point number with {DECIMALS} places.
+        ///      The rollover fee percentage returned by the fee policy will be no lower than
+        ///      this specified value.
```
brandoniles, line 102: Maybe add a comment that this effectively acts as a maximum debasement amount (when negative).

brandoniles (approved):
LGTM

PR #241, opened 2025-03-04
DR based flash swap fees

aalavandhan:
Flash swap fees, based on a piecewise linear function on dr before and after swap


spot-contracts/contracts/FeePolicy.sol
```diff
@@ -77,13 +79,9 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
-    /// @notice The lower bound of deviation ratio, below which some operations (which decrease the dr) are disabled.
-    uint256 public deviationRatioBoundLower;
-
-    /// @notice The upper bound of deviation ratio, above which some operations (which increase the dr) are disabled.
-    uint256 public deviationRatioBoundUpper;
-
-    //-----------------------------------------------------------------------------
+    /// @notice The lower and upper deviation ratio bounds outside which
+    ///         some operations (like flash swaps) which move dr away from 1.0 are disabled.
+    Range public drHardBounds;
```
aalavandhan, line 84: Making 2 uints into a struct with 2 uints should be storage safe. But we should test this on a mainnet fork before deploying.

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -128,11 +124,11 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
     /// @notice The percentage fee charged on burning vault notes.
     uint256 public vaultBurnFeePerc;
 
-    /// @notice The percentage fee charged by the vault to swap underlying tokens for perp tokens.
-    uint256 public vaultUnderlyingToPerpSwapFeePerc;
+    /// @notice Lower and upper fee percentages for flash minting.
+    Range public flashMintFeePercs;
```
aalavandhan, line 128: Same here. This should also be upgrade safe

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -78,13 +80,9 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
-    /// @notice The lower bound of deviation ratio, below which some operations (which decrease the dr) are disabled.
-    uint256 public deviationRatioBoundLower;
-
-    /// @notice The upper bound of deviation ratio, above which some operations (which increase the dr) are disabled.
-    uint256 public deviationRatioBoundUpper;
-
-    //-----------------------------------------------------------------------------
+    /// @notice The lower and upper deviation ratio bounds outside which
+    ///         some operations (like flash swaps) which move dr away from 1.0 are disabled.
+    Range public drHardBounds;
```
aalavandhan, line 85: We'll be deploying a new FeePolicy given significant upcoming changes, so we can safely ignore any changes to the storage layout

brandoniles: Combining these two properties into a struct is cleaner from a programming perspective, but I feel like it will make updating these values in the setter harder.

For example, when constructing a governance payload. Still worth it?

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -307,23 +299,55 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
         // swapping is disabled. (fees are set to 100%)
-        return (drPost < deviationRatioBoundLower ? ONE : vaultUnderlyingToPerpSwapFeePerc);
+        if (drPost <= drHardBounds.lower) {
+            return ONE;
+        }
+        return
+            LineHelpers
+                .computePiecewiseAvgY(
+                    Line({ x1: drHardBounds.lower, y1: flashMintFeePercs.upper, x2: ONE, y2: flashMintFeePercs.lower }),
+                    Line({ x1: ONE, y1: flashMintFeePercs.lower, x2: drHardBounds.upper, y2: flashMintFeePercs.lower }),
```
brandoniles, line 314: At first read, I thought the second `flashMintFeePercs.lower` should have been upper. But as-is means the line is flat on the other side of DR.

Looks coorect, but just verifying.

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -307,23 +299,55 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
+                    Range({ lower: drPost, upper: drPre }),
+                    ONE
+                )
+                .toUint256();
```
brandoniles, line 318: Should we have this function just return a uint, like in the original version?

brandoniles (commented):
Looks good, nice that it's simple

spot-contracts/contracts/FeePolicy.sol
aalavandhan: Yea, haven't really any issue with complex struct payloads.

spot-contracts/contracts/FeePolicy.sol
aalavandhan: Right. Now the breakpoint is dr = 1. In the next PR, we'll add `drSoftBounds` (similar to billy) which will allow us to configure the breakpoint ..

spot-contracts/contracts/FeePolicy.sol
aalavandhan: you mean `computePiecewiseAvgY`? I think that always returned an integer 

https://github.com/ampleforth/spot/blob/development/spot-vaults/contracts/_utils/LineHelpers.sol

PR #243, opened 2025-03-08
Restructuring perp/vault fees

aalavandhan:
1) Perp's mint/burn fees are paid to the vault (by minting/transfering perp tokens)
2) All fees collected during flash swaps go to the vault's balance. (before it was split between perp and the vault)
3) Added owner defined drSoftBounds (similar to bill broker), which determine when flash swap fees transition from constant value to the linearly increasing fee (previously the kink point was dr=1.0).

### DIFF
--- a/spot-contracts/contracts/RolloverVault.sol
+++ b/spot-contracts/contracts/RolloverVault.sol
@@ -4,7 +4,6 @@ pragma solidity ^0.8.20;
 import { IERC20Upgradeable, IPerpetualTranche, IBondController, ITranche, IFeePolicy } from "./_interfaces/IPerpetualTranche.sol";
 import { IVault } from "./_interfaces/IVault.sol";
 import { IRolloverVault } from "./_interfaces/IRolloverVault.sol";
-import { IERC20Burnable } from "./_interfaces/IERC20Burnable.sol";
 import { TokenAmount, RolloverData, SubscriptionParams } from "./_interfaces/CommonTypes.sol";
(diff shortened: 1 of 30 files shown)


spot-contracts/contracts/RolloverVault.sol
```diff
@@ -451,9 +450,7 @@ contract RolloverVault is
         IPerpetualTranche perp_ = perp;
         IERC20Upgradeable underlying_ = underlying;
         uint256 underlyingBalPre = underlying_.balanceOf(address(this));
-        (uint256 perpAmtOut, uint256 perpFeeAmtToBurn, SubscriptionParams memory s) = computeUnderlyingToPerpSwapAmt(
-            underlyingAmtIn
-        );
+        (uint256 perpAmtOut, , SubscriptionParams memory s) = computeUnderlyingToPerpSwapAmt(underlyingAmtIn);
```
brandoniles, line 453: Should we update `computeUnderlyingToPerpSwapAmt` to not return that second argument anymore? It doesn't have a use anymore I don't think.

brandoniles (commented):
I think we're good here, but do we have everything needed for the frontend to be able to represent fee activity?

spot-contracts/contracts/RolloverVault.sol
aalavandhan: Yea that was my first inclination as well. But considering some cowswap relayers have already integrated the current ABI, I'm leaning towards not rocking the boat.

aalavandhan:
> I think we're good here, but do we have everything needed for the frontend to be able to represent fee activity?

The mint/redeem preview functions should compute the fees and adjust for it. The UI should pick that up I think ..

brandoniles (approved):
LGTM

PR #246, opened 2025-03-11
Removed rollover fees in favor of daily rebalance

aalavandhan:
- Removed rollover fees entirely. 
- A new "rebalance" operation which can be executed at most once a day is responsible for direct value transfer between perp and the vault.
- When perp needs to be debased, we simply mint the vault more perp tokens.
- When perp needs to be enriched, we transfer some of the vault's underlying tokens into the perp reserve. (Which will then get rolled over into senior tranches through the normal ops).


spot-contracts/contracts/RolloverVault.sol
```diff
@@ -709,6 +748,46 @@ contract RolloverVault is
...
+        // such that it's value reflects in the vault's TVL,
+        // before the rebalance calculation.
+        _meldPerps(perp_);
+
+        SubscriptionParams memory s = _querySubscriptionState(perp_);
+        RebalanceData memory r = feePolicy.computeRebalanceData(s);
+        if (r.perpDebasement) {
+            // We transfer value from perp to the vault, by minting the vault perp tokens.
+            uint256 perpSupply = perp_.totalSupply();
+            uint256 perpAmtToVault = perpSupply.mulDiv(r.underlyingAmtToTransfer, s.perpTVL);
```
brandoniles, line 766:
```suggestion
            uint256 perpAmtToVault = r.underlyingAmtToTransfer.mulDiv(perpSupply, s.perpTVL);
```
A bit easier to understand, like a unit convertion, I think

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -709,6 +748,46 @@ contract RolloverVault is
...
+            perp_.debase(perpAmtToVault);
+            if (r.protocolFeeUnderlyingAmt > 0) {
+                // NOTE: We first mint the vault perp tokens, and then pay the protocol fee.
+                uint256 protocolFeePerpAmt = perpSupply.mulDiv(r.protocolFeeUnderlyingAmt, s.perpTVL);
```
brandoniles, line 770:
```suggestion
                uint256 protocolFeePerpAmt = r.protocolFeeUnderlyingAmt.mulDiv(perpSupply, s.perpTVL);
```

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -709,6 +748,46 @@ contract RolloverVault is
...
+                perp_.debase(protocolFeePerpAmt);
```
brandoniles, line 771: If we do one debase on the sum of (perpAmtToVault + r.protocolFeeUnderlyingAmt) above, it could save some gas.

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -709,6 +748,46 @@ contract RolloverVault is
...
+                IERC20Upgradeable(address(perp_)).safeTransfer(owner(), protocolFeePerpAmt);
```
brandoniles, line 772: As it's structured now, the protocol pays fees in spot in one case and ampl in the other. Since the fee is already denominated in ampl, I think it makes sense to keep it simple and just pay in ampl both ways.

We could move one meld operation above, then one conditional ampl transfer after.

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -461,6 +469,13 @@ contract PerpetualTranche is
         return r;
     }
 
+    /// @inheritdoc IPerpetualTranche
+    /// @dev CAUTION: Only the whitelisted vault can call this function. The logic controlling the frequency
+    ///      and magnitude of debasement should be thoroughly vetted.
+    function debase(uint256 perpAmtMint) external override onlyVault afterStateUpdate nonReentrant whenNotPaused {
```
brandoniles, line 475: A better alignment with interface / component ownership would be to have this function take in an `underlying` amount that should be transferred. Let perp be the authority for how it makes that happen.

Instead of a function called debase, I'd rather keep it neutral (inline with the two-sided view of the system comment). We could call this "balance" or "rebalance to vault", or similar. But probably best to come back to this after we figure out the other items.

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -74,6 +69,9 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
     /// @notice Target subscription ratio higher bound, 2.0 or 200%.
     uint256 public constant TARGET_SR_UPPER_BOUND = 2 * ONE;
 
+    /// @notice The enforced maximum change in perp's value per rebalance operation.
+    uint256 public constant MAX_PERP_VALUE_CHANGE_PERC = ONE / 100; // 0.01 or 1%
```
brandoniles, line 73: Do we need this? Could it get in the way in the future?

brandoniles (commented):
If we take more of a two-sided perspective on rebalancing, then I like the idea of the FeePolicy being the entry point for it, rather than the vault. It would be easy to keep symmetry that way too, as each component would have a "rebalance" function that's called by that policy.

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -57,6 +59,12 @@ import { BondHelpers } from "./_utils/BondHelpers.sol";
  *      When computing the value of assets in the system, the code always over-values by
  *      rounding up. When computing the value of incoming assets, the code rounds down.
  *
+ * @dev With release v5.0, we remove rollover fees entirely. Demand imbalance between perp and the vault
```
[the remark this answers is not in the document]
aalavandhan: yea we can get rid of the version mention

spot-contracts/contracts/RolloverVault.sol
aalavandhan: Yea that is cleaner

spot-contracts/contracts/FeePolicy.sol
aalavandhan: Yea lets get rid of this

spot-contracts/contracts/_interfaces/CommonTypes.sol
```diff
@@ -27,6 +27,15 @@ struct RolloverData {
     uint256 trancheInAmt;
 }
 
+struct RebalanceData {
+    /// @notice True, when value flows from perp into the vault and False when value flows from the vault into perp.
+    bool perpDebasement;
```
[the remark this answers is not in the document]
aalavandhan: I like that

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -709,6 +748,46 @@ contract RolloverVault is
     //--------------------------------------------------------------------------
     // Private write methods
 
+    /// @dev Executes a system-level rebalance. This operation transfers value between
+    ///      the perp reserve and the vault such that the system moves toward balance.
+    ///      Settles protocol fees, charged as a percentage of the bi-directional flow in value.
+    ///      Performs some book-keeping to keep track of the vault and perp's assets.
+    function _rebalance(IPerpetualTranche perp_, IERC20Upgradeable underlying_) private {
```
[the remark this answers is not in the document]
aalavandhan: Within the function the read is cached. If we don't pass it, there will be one additional storage read per variable in each function ...

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -273,6 +301,16 @@ contract RolloverVault is
     //--------------------------------------------------------------------------
     // External & Public write methods
 
+    /// @inheritdoc IRolloverVault
+    /// @dev This method can execute at-most once every 24 hours. The rebalance frequency is hard-coded and can't be changed.
+    function rebalance() external override nonReentrant whenNotPaused {
+        if (block.timestamp <= lastRebalanceTimestampSec + DAY_SEC) {
```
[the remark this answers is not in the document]
aalavandhan: We can make it configurable. Will add a todo for now.

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -379,9 +361,49 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
+        underlyingAmtToTransfer -= r.protocolFeeUnderlyingAmt;
+        r.underlyingAmtIntoPerp = (perpDebasement ? int256(-1) : int256(1)) * underlyingAmtToTransfer.toInt256();
+    }
+
+    /// @notice Computes the percentage change in perpTVL (to or from the vault) to bring the system to equilibrium.
+    /// @param s The subscription parameters of both the perp and vault systems.
+    /// @return debasement True, when value has to flow from perp into the vault and False when value has to flow from the vault into perp.
+    /// @return valueChangePerc The percentage of the perpTVL that has to flow out of perp (into the vault) or into perp (from the vault)
+    ///                         to make the system perfectly balanced (to make dr = 1.0).
+    function computePerpEquilibriumPerc(
```
brandoniles, line 392: Likely easier to skip calculating percentages, since this is a helper to a function that's only calculating raw balances to transfer anyway.

1) We have combined TVL
2) calc dr-normalized A (or Z)
3) figure out dr-normalized share of total TVL
4) Find difference between above and actual

aalavandhan: Done

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -709,6 +751,46 @@ contract RolloverVault is
...
+        // before the rebalance calculation.
+        _meldPerps(perp_);
+
+        SubscriptionParams memory s = _querySubscriptionState(perp_);
+        RebalanceData memory r = feePolicy.computeRebalanceData(s);
+        if (r.underlyingAmtIntoPerp <= 0) {
+            // We transfer value from perp to the vault, by minting the vault perp tokens.
+            // NOTE: We first mint the vault perp tokens, and then pay the protocol fee.
+            uint256 perpSupply = perp_.totalSupply();
+            uint256 perpAmtToVault = r.underlyingAmtIntoPerp.abs().mulDiv(perpSupply, s.perpTVL);
```
brandoniles, line 770: I still think it's better design to structure it like:
perp.debase(underlyingAmt)
or even better naming-- rebalanceToVault(underlyingAmt).

Since we're starting with underlying amounts here anyway.

Let perp handle the perp math and how the value transfer is implemented. Let the vault not have to worry about it.

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -136,6 +108,21 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
     /// @notice Lower and upper fee percentages for flash redemption.
     Range public flashRedeemFeePercs;
 
+    //-----------------------------------------------------------------------------
+    // Incentive parameters
+
+    /// @notice The magnitude of perp's daily debasement (when dr <= 1).
+    uint256 public debasementValueChangePerc;
```
brandoniles, line 115: These are maximum values, right?

How about naming like--
`maxPerpDebasePerc` and `maxPerpEnrichPerc`

(But see the other comment about whether we need these or not)

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -379,9 +365,44 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
+            (s.seniorTR * ONE) + (juniorTR * targetSubscriptionRatio)
+        );
+
+        uint256 reqPerpTVL = (s.perpTVL + s.vaultTVL).mulDiv(drNormalizedSeniorTR, ONE);
+        bool perpDebasement = reqPerpTVL <= s.perpTVL;
+        uint256 perpValueDelta = 0;
+        if (perpDebasement) {
+            perpValueDelta = MathUpgradeable.min(
+                s.perpTVL - reqPerpTVL,
+                s.perpTVL.mulDiv(debasementValueChangePerc, ONE)
```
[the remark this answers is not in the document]
aalavandhan: Well there's some point where we can conclude that the change isn't worth making.. For example, 1 say AMPL? 

Currently It just set to 0.01% of the supply. Mostly to guard against dust rebalancing. Alternatively we could make the configurable like AMPL's equilibrium band.

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -709,6 +751,46 @@ contract RolloverVault is
...
 
+    /// @dev Executes a system-level rebalance. This operation transfers value between
+    ///      the perp reserve and the vault such that the system moves toward balance.
+    ///      Settles protocol fees, charged as a percentage of the bi-directional flow in value.
+    ///      Performs some book-keeping to keep track of the vault and perp's assets.
+    function _rebalance(IPerpetualTranche perp_, IERC20Upgradeable underlying_) private {
+        // If the vault has any perp tokens we deconstruct it
+        // such that it's value reflects in the vault's TVL,
+        // before the rebalance calculation.
+        _meldPerps(perp_);
```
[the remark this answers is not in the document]
aalavandhan: Meld should be no-op if there are no perps in the vault. So we can call it many times and it should be okay. 

Voltroning tranches doesn't affect the tvl math, however breaking perps apart does. So yea, full recover is probably wasteful here.

spot-contracts/contracts/FeePolicy.sol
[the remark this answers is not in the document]
aalavandhan: If perp supply needs to shrink by 50%, we don't really want to do it all at once? We'll do a change at most as big as `debasementValueChangePerc` or `enrichmentValueChangePerc` every time.

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -379,9 +365,44 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
+    function computeRebalanceData(SubscriptionParams memory s) external view override returns (RebalanceData memory r) {
+        uint256 juniorTR = (TRANCHE_RATIO_GRANULARITY - s.seniorTR);
+        uint256 drNormalizedSeniorTR = ONE.mulDiv(
+            (s.seniorTR * ONE),
+            (s.seniorTR * ONE) + (juniorTR * targetSubscriptionRatio)
+        );
+
+        uint256 reqPerpTVL = (s.perpTVL + s.vaultTVL).mulDiv(drNormalizedSeniorTR, ONE);
+        bool perpDebasement = reqPerpTVL <= s.perpTVL;
+        uint256 perpValueDelta = 0;
```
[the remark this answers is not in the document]
aalavandhan: we'll have to go back and forth between signed and unsigned .. but I guess its fine

spot-contracts/contracts/RolloverVault.sol
aalavandhan: Yes that's simpler

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -111,11 +113,11 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
     //-----------------------------------------------------------------------------
     // Incentive parameters
 
-    /// @notice The magnitude of perp's daily debasement (when dr <= 1).
-    uint256 public debasementValueChangePerc;
+    /// @notice The maximum allowed daily perp debasement percentage (when dr <= 1).
```
brandoniles, line 116: Make these max values a percentage of combined TVL (SPOT + stAMPL), rather than percentage of just one side

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -245,12 +246,15 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
-    /// @notice Updates the rebalance reaction lag.
-    /// @param maxPerpDebasePerc_ The magnitude of the daily debasement.
-    /// @param maxPerpEnrichPerc_ The magnitude of the daily enrichment.
-    function updateRebalanceRate(uint256 maxPerpDebasePerc_, uint256 maxPerpEnrichPerc_) external onlyOwner {
-        maxPerpDebasePerc = maxPerpDebasePerc_;
-        maxPerpEnrichPerc = maxPerpEnrichPerc_;
+    /// @notice Updates the rebalance rate.
+    /// @param debasementSystemTVLPerc_ The percentage of system tvl out of perp on debasement.
+    /// @param enrichmentSystemTVLPerc_ The percentage of system tvl into perp on enrichment.
+    function updateMaxRebalancePerc(
```
brandoniles, line 252: `updateRebalanceRates` is probably best, given the names moved away from the "max" terminology. We can keep these inline.

aalavandhan: done 👍

PR #247, opened 2025-03-14
Pair operations

aalavandhan:
Mint2 & Redeem2 => Ability to mint/redeem spot and stampl as matching pairs.

### DIFF
--- a/spot-contracts/contracts/RolloverVault.sol
+++ b/spot-contracts/contracts/RolloverVault.sol
@@ -18,6 +18,7 @@ import { BondTranches, BondTranchesHelpers } from "./_utils/BondTranchesHelpers.
 import { TrancheHelpers } from "./_utils/TrancheHelpers.sol";
 import { BondHelpers } from "./_utils/BondHelpers.sol";
 import { PerpHelpers } from "./_utils/PerpHelpers.sol";
+import { ERC20Helpers } from "./_utils/ERC20Helpers.sol";
 import { TrancheManager } from "./_utils/TrancheManager.sol";
 
 /*
@@ -68,6 +69,7 @@ contract RolloverVault is
 
     // ERC20 operations
     using SafeERC20Upgradeable for IERC20Upgradeable;
+    using ERC20Helpers for IERC20Upgradeable;
 
     // math
     using MathUpgradeable for uint256;
@@ -429,19 +431,19 @@ contract RolloverVault is
     /// @dev This operation pushes the system back into balance, we thus charge no fees.
     function mint2(
         uint256 underlyingAmtIn
-    ) external override nonReentrant whenNotPaused returns (uint256 perpAmt, uint256 noteAmt) {
+    ) external override nonReentrant whenNotPaused returns (uint256 perpAmt, uint256 vaultNoteAmt) {
         // Compute perp vault split
         IPerpetualTranche perp_ = perp;
         IERC20Upgradeable underlying_ = underlying;
         SubscriptionParams memory s = _querySubscriptionState(perp_);
-        (uint256 underlyingAmtIntoPerp, uint256 underlyingAmtIntoVault) = feePolicy.computeNeutralSplit(
+        (uint256 underlyingAmtIntoPerp, uint256 underlyingAmtIntoVault) = feePolicy.computeDREquilibriumSplit(
             underlyingAmtIn,
             s.seniorTR
         );
 
-        // Compute perp amount and note amount to mint
+        // Compute perp amount and vault note amount to mint
         perpAmt = perp_.totalSupply().mulDiv(underlyingAmtIntoPerp, s.perpTVL);
-        noteAmt = computeMintAmt(underlyingAmtIntoVault, 0);
+        vaultNoteAmt = computeMintAmt(underlyingAmtIntoVault, 0);
 
         // Transfer underlying tokens from user
         underlying_.safeTransferFrom(msg.sender, address(this), underlyingAmtIn);
@@ -451,7 +453,7 @@ contract RolloverVault is
         IERC20Upgradeable(address(perp_)).safeTransfer(msg.sender, perpAmt);
 
         // Mint vault notes to user
-        _mint(msg.sender, noteAmt);
+        _mint(msg.sender, vaultNoteAmt);
 
         // Sync underlying
         _syncAsset(underlying_);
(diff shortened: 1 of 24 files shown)


spot-contracts/contracts/RolloverVault.sol
```diff
@@ -624,61 +716,64 @@ contract RolloverVault is
...
-    function computeMintAmt(uint256 underlyingAmtIn) public view returns (uint256) {
-        //-----------------------------------------------------------------------------
-        uint256 feePerc = feePolicy.computeVaultMintFeePerc();
-        //-----------------------------------------------------------------------------
+    /// @notice Computes the amount of notes minted when given amount of underlying asset tokens
+    ///         are deposited into the system.
+    /// @param underlyingAmtIn The amount underlying tokens to be deposited into the vault.
+    /// @param feePerc The percentage of minted notes paid as fees.
+    /// @return noteAmtMint The amount of notes to be minted.
+    function computeMintAmt(uint256 underlyingAmtIn, uint256 feePerc) public view returns (uint256 noteAmtMint) {
```
brandoniles, line 724: What's the motivation for replacing the line
`uint256 feePerc = feePolicy.computeVaultMintFeePerc();`
with the input parameter `feePerc`?

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -624,61 +716,64 @@ contract RolloverVault is
...
-        notes = notes.mulDiv(ONE - feePerc, ONE);
-        return notes;
+        noteAmtMint = noteAmtMint.mulDiv(ONE - feePerc, ONE);
     }
 
-    /// @inheritdoc IVault
-    function computeRedemptionAmts(uint256 noteAmtBurnt) public view returns (TokenAmount[] memory) {
+    /// @notice Computes the amount of asset tokens redeemed when burning given number of vault notes.
+    /// @param noteAmtBurnt The amount of notes to be burnt.
+    /// @param feePerc The percentage of minted notes paid as fees.
```
brandoniles, line 739:
```suggestion
    /// @param feePerc The percentage of redeemed notes paid as fees.
```
But pay attention to the other comment about the semantic difference between "redeem" and "burn", so we're consistent.

spot-contracts/contracts/_utils/TrancheManager.sol
```diff
@@ -0,0 +1,85 @@
...
+    using SafeERC20Upgradeable for IERC20Upgradeable;
+
+    // math
+    using MathUpgradeable for uint256;
+
+    //--------------------------------------------------------------------------
+    // Helper methods
+
+    /// @notice Low level method which tranches underlying tokens into the provided bond contract.
+    function execTranche(IBondController bond, IERC20Upgradeable underlying_, uint256 underlyingAmt) public {
```
brandoniles, line 38: Should all these public function be external?

Assuming we don't want internal, because we want to decrease the bytecode size of RolloverVault.

brandoniles, line 38: This function is maybe better placed in BondHelpers.sol? Seems like a library function that would be called with the `using` pattern. It already takes the bond as the first argument.

spot-contracts/contracts/_utils/TrancheManager.sol
```diff
@@ -0,0 +1,85 @@
...
+            bond.mature();
+        }
+        bond.redeemMature(address(tranche), amount);
+    }
+
+    /// @notice Low level method that redeems the given tranche for the underlying asset, before maturity.
+    ///         If the contract holds sibling tranches with proportional balances, those will also get redeemed.
+    ///         It interacts with the button-wood bond contract.
+    function execImmatureTrancheRedemption(IBondController bond, BondTranches memory bt) public {
+        uint256[] memory trancheAmts = bt.computeRedeemableTrancheAmounts(address(this));
```
brandoniles, line 57: Does using `this` in this context actually work?

A library function that's marked as `internal` is included in the bytecode of the contract that's using it. But external library functions are deployed as their own separate contract. So, I would expect `this` to no longer be the address of the RolloverVault that holds tranches.

spot-contracts/contracts/_utils/TrancheManager.sol
```diff
@@ -0,0 +1,85 @@
...
+
+        // NOTE: It is guaranteed that if one tranche amount is zero, all amounts are zeros.
+        if (trancheAmts[0] > 0) {
+            bond.redeem(trancheAmts);
+        }
+    }
+
+    /// @notice Checks if the spender has sufficient allowance. If not, approves the maximum possible amount.
+    function checkAndApproveMax(IERC20Upgradeable token, address spender, uint256 amount) public {
```
brandoniles, line 66: Seems weird to have this function in a library called TrancheManager. Is there a better place for it? Seems like an ERC20 helper.

(Same comment for use of `this`)

brandoniles (commented):
I'd like to talk about the ratio computation strategy before digging into the mint2 & redeem2 functions themselves. But here are some comments from an initial pass.

On naming:
Using the term "note" to describe the vault is ambiguous, since perp is technically also a note. We should call the rotation vault notes something more specific, like maybe "stAMPL" or "rolloverVaultNotes".

Then see also the difference between redemption and burning.

spot-contracts/contracts/_interfaces/IFeePolicy.sol
```diff
@@ -45,4 +45,28 @@ interface IFeePolicy {
     /// @return r Rebalance data, magnitude and direction of value flow between perp and the rollover vault
     ///           expressed in the underlying token amount and the protocol's cut.
     function computeRebalanceData(SubscriptionParams memory s) external view returns (RebalanceData memory r);
+
+    /// @notice Computes the dr-neutral split of underlying tokens into perp and the vault.
```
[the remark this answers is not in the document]
aalavandhan: yea I think that's right. The function names are flipped ..

spot-contracts/contracts/_utils/TrancheManager.sol
aalavandhan: But this is an external / linked library which is an external contract called using `delegatecall`.  In which case `this` should point to the caller contract?

spot-contracts/contracts/_utils/TrancheManager.sol
aalavandhan: Sure they can all be external

aalavandhan: Structurally that makes sense, but moving it there wont help trim the bytecode

spot-contracts/contracts/_interfaces/IRolloverVault.sol
```diff
@@ -2,13 +2,30 @@
...
 interface IRolloverVault is IVault {
     /// @notice Gradually transfers value between the perp and vault, to bring the system back into balance.
     /// @dev The rebalance function can be executed at-most once a day.
     function rebalance() external;
 
+    /// @notice Batch operation to mint both perp and rollover vault tokens.
+    /// @param underlyingAmtIn The amount of underlying tokens to be tranched.
+    /// @return perpAmt The amount of perp tokens minted.
+    /// @return noteAmt The amount of vault notes minted.
+    function mint2(uint256 underlyingAmtIn) external returns (uint256 perpAmt, uint256 noteAmt);
```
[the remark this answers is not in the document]
aalavandhan: hmm. That might not be generic enough for this interface. Maybe we could say perpNoteAmt, vaultNoteAmt ..

spot-contracts/contracts/_interfaces/IRolloverVault.sol
```diff
@@ -2,13 +2,30 @@
...
+
+    /// @notice Batch operation to burn both perp and rollover vault tokens for the underlying collateral and tranches.
+    /// @param perpAmtAvailable The amount of perp tokens available to burn.
```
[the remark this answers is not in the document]
aalavandhan: Right, we might have to do a system-wide pass and make this consistent.

spot-contracts/contracts/_utils/TrancheManager.sol
aalavandhan: Okay we can move it to a erc20 helpers lib file

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -421,38 +427,126 @@ contract RolloverVault is
...
+
+        // Mint perps to user
+        _trancheAndMintPerps(perp_, underlying_, s.perpTVL, s.seniorTR, perpAmt);
+        IERC20Upgradeable(address(perp_)).safeTransfer(msg.sender, perpAmt);
+
+        // Mint vault notes to user
+        _mint(msg.sender, vaultNoteAmt);
+
+        // Sync underlying
+        _syncAsset(underlying_);
```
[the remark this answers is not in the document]
aalavandhan: That should happen with `_trancheAndMintPerps`

brandoniles (approved):
LGTM

PR #249, opened 2025-04-01
Rebalancing updates

aalavandhan:
1) Configurable equilibrium zone rebalancing
2) Configurable rebalancing frequency
3) While transferring value from the vault to perps, we transfer tranches rather than AMPL

### DIFF
--- a/spot-contracts/contracts/FeePolicy.sol
+++ b/spot-contracts/contracts/FeePolicy.sol
@@ -257,10 +257,10 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
         flashRedeemFeePercs = flashRedeemFeePercs_;
     }
 
-    /// @notice Updates the rebalance rate.
+    /// @notice Updates the rebalance rates.
     /// @param debasementSystemTVLPerc_ The percentage of system tvl out of perp on debasement.
     /// @param enrichmentSystemTVLPerc_ The percentage of system tvl into perp on enrichment.
-    function updateMaxRebalancePerc(
+    function updateRebalanceRates(
         uint256 debasementSystemTVLPerc_,
         uint256 enrichmentSystemTVLPerc_
     ) external onlyOwner {
--- a/spot-contracts/test/FeePolicy.ts
+++ b/spot-contracts/test/FeePolicy.ts
@@ -313,18 +313,18 @@ describe("FeePolicy", function () {
     });
   });
 
-  describe("#updateMaxRebalancePerc", function () {
+  describe("#updateRebalanceRates", function () {
     describe("when triggered by non-owner", function () {
       it("should revert", async function () {
         await expect(
-          feePolicy.connect(otherUser).updateMaxRebalancePerc(toPerc("0.005"), toPerc("0.01")),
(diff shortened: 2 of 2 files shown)


spot-contracts/contracts/RolloverVault.sol
```diff
@@ -259,6 +262,12 @@ contract RolloverVault is
         keeper = keeper_;
     }
 
+    /// @notice Updates the rebalance frequency.
+    /// @param rebalanceFreqSec_ The new rebalance frequency in seconds.
+    function updateRebalanceFrequency(uint256 rebalanceFreqSec_) public onlyOwner {
```
brandoniles, line 267:
```suggestion
    function updateRebalanceFrequency(uint256 rebalanceFreqSec_) external onlyOwner {
```

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -376,6 +387,12 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
 
     /// @inheritdoc IFeePolicy
     function computeRebalanceData(SubscriptionParams memory s) external view override returns (RebalanceData memory r) {
+        // We skip rebalancing if dr is close to 1.0
+        uint256 dr = computeDeviationRatio(s);
+        if (dr > rebalEqDr.lower && dr < rebalEqDr.upper) {
```
brandoniles, line 392:
```suggestion
        if (dr >= rebalEqDr.lower && dr <= rebalEqDr.upper) {
```
Shouldn't really matter in practice, but I find it easier to think of this as a closed interval.

spot-contracts/contracts/FeePolicy.sol
```diff
@@ -400,12 +417,6 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
             );
         }
 
-        // We skip "dust" value transfer
-        uint256 minPerpAbsValueDelta = s.perpTVL.mulDiv(EQUILIBRIUM_REBALANCE_PERC, ONE);
```
brandoniles, line 404: Is it possible for this check to still be relevant, even when there's an equilibrium range?

For example, DR is out of range, but also the amount of liquid AMPL in the vault is very low. So a rebalance would still trigger on a small amount.

aalavandhan: Well it can't though if the dr range is wide enough.. A dr range of [0.98, 1.02], would mean ~2% perp tvl perc delta?

brandoniles: Say the dr equilibrium range is [0.98, 1.02], and the DR is at 0.5. There's 250,001 liquid AMPL in the rotation vault, and it reserves 250K AMPL that it doesn't use to rotate. That means the DR is out of range, but the rotation size would be 1 AMPL. This could be considered dust

aalavandhan: ah. Good point .. yes

aalavandhan: The liquidity thresholds just limit rollovers though. This is related to rebalancing?

brandoniles (approved):
LGTM

PR #252, opened 2025-04-23
Dev Audit Changes

aalavandhan:
**Major changes:**
1) Fee policy overhaul. DR based unified fee function. New rebalancing math and frequency control.
2) Perp collects mint/burn fees and pays the vault once during rebalance.
3) Updated perp/vault fees and rebalance handling.

### DIFF
--- a/spot-contracts/contracts/FeePolicy.sol
+++ b/spot-contracts/contracts/FeePolicy.sol
@@ -3,7 +3,7 @@ pragma solidity ^0.8.20;
 
 import { IFeePolicy } from "./_interfaces/IFeePolicy.sol";
 import { SubscriptionParams, Range, Line } from "./_interfaces/CommonTypes.sol";
-import { InvalidPerc, InvalidFees } from "./_interfaces/ProtocolErrors.sol";
+import { InvalidPerc, InvalidRange, InvalidFees } from "./_interfaces/ProtocolErrors.sol";
 
 import { LineHelpers } from "./_utils/LineHelpers.sol";
 import { MathUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/math/MathUpgradeable.sol";
@@ -170,13 +170,15 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
         // Expect fees to be non-decreasing when dr moves away from 1.0
         validFees = ((feeFnDRDown_.y1 >= feeFnDRDown_.y2) && (feeFnDRUp_.y1 <= feeFnDRUp_.y2)) && validFees;
 
+        // Expect fee percentages to be valid
+        validFees =
+            (feeFnDRDown_.y1 <= ONE) &&
+            (feeFnDRDown_.y2 <= ONE) &&
+            (feeFnDRUp_.y1 <= ONE) &&
+            ((feeFnDRUp_.y2 <= ONE) && validFees);
+
         if (!validFees) {
             revert InvalidFees();
-        }
-
-        // Expect fee percentages to be valid
-        if ((feeFnDRDown_.y1 > ONE) || (feeFnDRDown_.y2 > ONE) || (feeFnDRUp_.y1 > ONE) || (feeFnDRUp_.y2 > ONE)) {
-            revert InvalidPerc();
         }
 
         feeFnDRDown = feeFnDRDown_;
@@ -196,6 +198,13 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
         Range memory perpEnrichmentPercLimits_,
         uint256 rebalanceFreqSec_
     ) external onlyOwner {
+        if (
+            perpDebasementPercLimits_.lower > perpDebasementPercLimits_.upper ||
+            perpEnrichmentPercLimits_.lower > perpEnrichmentPercLimits_.upper
+        ) {
+            revert InvalidRange();
+        }
+
         perpDebasementLag = perpDebasementLag_;
         perpEnrichmentLag = perpEnrichmentLag_;
         perpDebasementPercLimits = perpDebasementPercLimits_;
@@ -226,7 +235,7 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
                 LineHelpers
                     .computePiecewiseAvgY(
                         fee,
-                        Line({ x1: ONE, y1: fee.y2, x2: fee.x2, y2: fee.y2 }),
+                        Line({ x1: fee.x2, y1: fee.y2, x2: ONE, y2: fee.y2 }),
                         Range({ lower: drPost, upper: drPre }),
                         fee.x2
                     )
--- a/spot-contracts/contracts/FeePolicy.sol
+++ b/spot-contracts/contracts/FeePolicy.sol
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.20;
 
 import { IFeePolicy } from "./_interfaces/IFeePolicy.sol";
-import { SubscriptionParams, Range, Line } from "./_interfaces/CommonTypes.sol";
+import { SystemTVL, Range, Line } from "./_interfaces/CommonTypes.sol";
 import { InvalidPerc, InvalidRange, InvalidFees } from "./_interfaces/ProtocolErrors.sol";
 
 import { LineHelpers } from "./_utils/LineHelpers.sol";
@@ -22,21 +22,21 @@ import { MathHelpers } from "./_utils/MathHelpers.sol";
  *          supports rolling over all mature collateral backing perps.
  *
  *          The system's balance is defined by it's `deviationRatio` which is defined as follows.
- *              - `subscriptionRatio`   = (vaultTVL * seniorTR) / (perpTVL * 1-seniorTR)
- *              - `deviationRatio` (dr) = subscriptionRatio / targetSubscriptionRatio
+ *              - `systemRatio`   = vaultTVL / perpTVL
+ *              - `deviationRatio` (dr) = systemRatio / targetSystemRatio
  *
  *          When the dr = 1, the system is considered perfectly balanced.
- *          When the dr < 1, it's considered "under-subscribed".
- *          When the dr > 1, it's considered "over-subscribed".
+ *          When the dr < 1, the vault is considered "under-subscribed".
+ *          When the dr > 1, the vault is  considered "over-subscribed".
  *
  *          Fees:
  *          - The system charges a greater fee for operations that move it away from the balance point.
  *          - If an operation moves the system back to the balance point, it charges a lower fee (or no fee).
  *
  *          Incentives:
- *          - When the system is "under-subscribed", value is transferred from perp to the vault at a predefined rate.
+ *          - When the vault is "under-subscribed", value is transferred from perp to the vault at the computed rate.
  *            This debases perp tokens gradually and enriches the rollover vault.
- *          - When the system is "over-subscribed", value is transferred from the vault to perp at a predefined rate.
+ *          - When the vault is "over-subscribed", value is transferred from the vault to perp at the computed rate.
  *            This enriches perp tokens gradually and debases the rollover vault.
  *          - This transfer is implemented through a periodic "rebalance" operation, executed by the vault, and
  *            gradually nudges the system back into balance. On rebalance, the vault queries this policy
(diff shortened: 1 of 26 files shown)


spot-contracts/contracts/FeePolicy.sol
```diff
@@ -172,205 +154,97 @@ contract FeePolicy is IFeePolicy, OwnableUpgradeable {
...
-            return ONE;
+    function computeFeePerc(uint256 drPre, uint256 drPost) public view override returns (uint256) {
+        // DR is decreasing, we use feeFnDRDown
+        if (drPre > drPost) {
+            Line memory fee = feeFnDRDown;
+            return
+                LineHelpers
+                    .computePiecewiseAvgY(
+                        fee,
+                        Line({ x1: ONE, y1: fee.y2, x2: fee.x2, y2: fee.y2 }),
```
brandoniles, line 229: I expected it to go like:
`Line({ x1: fee.x2, y1: fee.y2, x2: ONE, y2: fee.y2 }),`

So we keep the pattern of going left to right

brandoniles (commented):
For first commit, fee policy restructure

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -470,12 +471,42 @@ contract PerpetualTranche is
...
+        uint256 protocolSharePerc = feePolicy.protocolSharePerc();
+        if (protocolSharePerc > 0) {
+            _mint(feePolicy.protocolFeeCollector(), protocolSharePerc.mulDiv(totalSupply(), ONE - protocolSharePerc));
+        }
+    }
+
+    /// @inheritdoc IPerpetualTranche
+    /// @dev Only the whitelisted vault can call this function.
     ///      The logic controlling the frequency and magnitude of debasement should be vetted.
+    ///      Pays a protocol fee on every rebalance.
```
brandoniles, line 495: Nothing in this function calls a protocol fee, so I think this comment is better placed somewhere else.

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -470,12 +471,42 @@ contract PerpetualTranche is
...
     function rebalanceToVault(
-        uint256 underlyingAmtToTransfer
+        int256 underlyingAmtToTransfer
     ) external override onlyVault afterStateUpdate nonReentrant whenNotPaused {
-        _mint(address(vault), underlyingAmtToTransfer.mulDiv(totalSupply(), _reserveValue()));
+        // When value is flowing out of perp to the vault, we mint the vault perp tokens.
+        if (underlyingAmtToTransfer < 0) {
```
brandoniles, line 500: This sign seems backwards to me. If we're "rebalancing to the vault" a positive value, wouldn't that mean that we're transferring value out?

We should also probably have a docstring for this param, if there's any confusion at all.

brandoniles (commented):
For commit #2

spot-contracts/contracts/PerpetualTranche.sol
```diff
@@ -470,12 +471,42 @@ contract PerpetualTranche is
...
+            perp_.safeTransfer(to, collectedBal);
+        }
+    }
+
+    /// @inheritdoc IPerpetualTranche
+    /// @dev Only the whitelisted vault can call this function.
+    function payProtocolFee() external override onlyVault afterStateUpdate nonReentrant whenNotPaused {
+        uint256 protocolSharePerc = feePolicy.protocolSharePerc();
+        if (protocolSharePerc > 0) {
+            _mint(feePolicy.protocolFeeCollector(), protocolSharePerc.mulDiv(totalSupply(), ONE - protocolSharePerc));
```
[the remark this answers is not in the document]
aalavandhan: If we want the protocol to take 1% of the existing tvl as fees, we have to mint 1.01% more notes.

spot-contracts/contracts/PerpetualTranche.sol
aalavandhan: hmm, then I think we can just make `underlyingAmtToTransfer` to unsigned. And on perp simply transfer value if it's non-zero.

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -482,14 +460,19 @@ contract RolloverVault is
...
         IPerpetualTranche perp_ = perp;
-        (perpAmt, vaultNoteAmt) = feePolicy.computeDRNeutralSplit(
-            perpAmtAvailable,
-            vaultNoteAmtAvailable,
-            perp_.totalSupply(),
-            totalSupply()
-        );
+
+        // Compute perp vault split
+        {
```
brandoniles, line 466: There are a few places where you added extra namespaces/nesting. It's a pattern that's useful in C++ for RAII, but don't think that's relevant here. Is there another reason? If it's for documentation, I'd rather do that in docstrings rather than something that could potentially impact the compiled output.

(and other locations)

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -1080,8 +1087,9 @@ contract RolloverVault is
     }
 
     /// @dev Logs the token balance held by the vault.
-    function _syncAsset(IERC20Upgradeable token) private {
-        emit AssetSynced(token, token.balanceOf(address(this)));
+    function _syncAsset(IERC20Upgradeable token) private returns (uint256 balance) {
```
brandoniles, line 1087: Of all the calls to _syncAsset, only one uses the balance return value. That call was already doing the read too.

Seems more efficient to keep it the way it was before, with no read and no return val? Otherwise, we're adding lots more reads and not saving on anything else.

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -861,34 +861,42 @@ contract RolloverVault is
...
-        } else {
-            // We transfer value from the vault to perp, by minting the perp tokens (after making required deposit)
-            // and then simply burning the newly minted perp tokens.
-            uint256 perpAmtToTransfer = r.underlyingAmtIntoPerp.toUint256().mulDiv(perp_.totalSupply(), s.perpTVL);
-            _trancheAndMintPerps(perp_, underlying_, s.perpTVL, s.seniorTR, perpAmtToTransfer);
-            IERC20Burnable(address(perp_)).burn(perpAmtToTransfer);
+        int256 underlyingAmtIntoPerp = feePolicy.computeRebalanceAmount(s);
+
+        // We rebalance from perp to the vault.
+        perp_.rebalanceToVault(underlyingAmtIntoPerp);
```
brandoniles: Should we put a conditional around this? I think it's safe to say that if amt into perp >= 0, we don't need to tell perp to send vault anything. It would also save us the meld operation.

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -861,34 +861,42 @@ contract RolloverVault is
...
+                IERC20Burnable(address(perp_)).burn(perpAmtToTransfer);
+            }
+            // otherwise, no value transfer here.
         }
 
-        // Pay protocol fees
-        if (r.protocolFeeUnderlyingAmt > 0) {
-            underlying_.safeTransfer(owner(), r.protocolFeeUnderlyingAmt);
+        // We pay the protocol fee on every rebalance.
+        perp.payProtocolFee();
```
brandoniles: I put this inside the conditional too

spot-contracts/contracts/RolloverVault.sol
```diff
@@ -861,34 +861,42 @@ contract RolloverVault is
...
+        {
+            uint256 protocolSharePerc = feePolicy.protocolSharePerc();
+            if (protocolSharePerc > 0) {
+                _mint(
+                    feePolicy.protocolFeeCollector(),
+                    protocolSharePerc.mulDiv(totalSupply(), ONE - protocolSharePerc)
```
brandoniles: Is the `ONE - protocolSharePerc` correct?

brandoniles (commented):
For Commit 3

spot-contracts/contracts/_interfaces/IFeePolicy.sol
```diff
@@ -41,34 +16,23 @@ interface IFeePolicy {
...
+    /// @notice Computes magnitude and direction of value flow between perp and the rollover vault
+    ///         expressed in underlying tokens.
     /// @param s The subscription parameters of both the perp and vault systems.
-    /// @return r Rebalance data, magnitude and direction of value flow between perp and the rollover vault
-    ///           expressed in the underlying token amount and the protocol's cut.
-    function computeRebalanceData(SubscriptionParams memory s) external view returns (RebalanceData memory r);
+    /// @return underlyingAmtIntoPerp The value in underlying tokens, that need to flow from the vault into perp.
+    function computeRebalanceAmount(SubscriptionParams memory s) external view returns (int256 underlyingAmtIntoPerp);
+
+    /// @return The share of the system tvl paid to the protocol owner as fees.
```
brandoniles, line 30: per rebalance period?

A reader could easily interpret this as an annual amount.

brandoniles (commented):
Commit #4

spot-contracts/contracts/RolloverVault.sol
aalavandhan: The compiler does limit the number of variables in a block to 16, If I remember correctly. So using blocks, helps keep the function namespace cleaner (ie ones declared in the block wont propagate up). I've used them from time to time specifically for that reason. But here, we are nowhere close to 16 variables.

spot-contracts/contracts/RolloverVault.sol
aalavandhan: We were running into contract size limits, so rolled this up to make some room.

spot-contracts/contracts/RolloverVault.sol
aalavandhan: Right

spot-contracts/contracts/RolloverVault.sol
aalavandhan: 👍🏽

spot-contracts/contracts/RolloverVault.sol
brandoniles: I don't see how this helps any, though

We move the line `uint256 balance = token.balanceOf(address(this));` from `_syncDeployedAsset` to here, so it doesn't remove any code.

But then it makes all the other calls to `_syncAsset` less efficient because there's an extra read in that call path now.

aalavandhan: Changing it back put the contract over the size limit. 
Screenshot 2025-05-06 at 12 20 26 PM.

Here is a explanation from chatGPT. Seems to make sense ..


**Explaining code size differences**

The first code snippet has a larger compiled size because it duplicates code: both `_syncDeployedAsset` and `_syncAsset` each call `token.balanceOf` and emit events. This repetition increases bytecode size. The second snippet reduces duplication by modifying `_syncAsset` to return the balance, with `_syncDeployedAsset` then calling the modified `_syncAsset`. This centralization cuts down on unnecessary code duplication, and since both functions are private, the calls are internal, minimizing the need for extra dispatch. This centralization leads to a more efficient bytecode.


Thought for 6 seconds


The reason boils down to **duplication of EVM bytecode** versus **re-use via an internal call**:

1. **First snippet duplicates the same sequence twice**

   * In `_syncDeployedAsset` you do:

     ```solidity
     uint256 balance = token.balanceOf(address(this));
     emit AssetSynced(token, balance);
     ```
   * And then in `_syncAsset` you do *exactly* the same two operations again.
     The Solidity compiler does *not* automatically merge identical code sequences across distinct functions. So you end up paying the size-cost of `CALL` (to invoke `balanceOf`) plus the `LOG` (opcode for the `emit`) twice in your final bytecode.

2. **Second snippet centralizes it behind one function**

   ```solidity
   function _syncAsset(IERC20Upgradeable token) private returns (uint256 balance) {
     balance = token.balanceOf(address(this));
     emit AssetSynced(token, balance);
   }
   function _syncDeployedAsset(IERC20Upgradeable token) private {
     uint256 balance = _syncAsset(token);
     // ...
   }
   ```

   Now the compiler only needs one copy of:

   * the external `balanceOf` call
   * the `AssetSynced` event log

   Instead of embedding them directly in both places, it emits a small internal‐call stub (a `JUMP` to `_syncAsset`) plus the single implementation. Internally, that looks roughly like:

   ```
   ... push token …
   JUMP to _syncAsset
   … // later in the bytecode
   JUMPDEST (_syncAsset):
     CALL balanceOf
     LOG AssetSynced
     PUSH1 0x40  // return the balance
     RETURN
   ```

   So you pay the overhead of one tiny `JUMP` instruction, but you *save* the full size of duplicating the `CALL` and `LOG` sequences a second time.

---

### Key takeaway

Whenever you have identical logic in more than one place, factor it out into an internal (or private) function that returns whatever you need. That way the compiler generates:

* **One** copy of the heavy-weight logic
* A **small** jump table entry for each caller

…instead of duplicating the full opcode sequence each time, which directly inflates your contract’s bytecode size.

brandoniles (approved):
LGTM

PR #254, opened 2025-05-04
Removed subscription ratio for a global vault:perp tvl (or system ratio)

aalavandhan:
Now the spot system doesn't care about the bond ratios used by the underlying tranches. It balances incentives and fees based on an owner defined vaultTVL : perpTVL target ratio (or system ratio)


brandoniles (commented):
Looks good. Just the one comment on naming.

brandoniles (approved):
LGTM

PR #258, opened 2025-06-06
Fee update

aalavandhan:
If we think the fees act as brakes, then we should be able to start applying the fees before the system gets out of balance.

So for example, when dr > 1; and dr is decreasing we can start applying fees at 1.05 (ie before we get to 1). Similarly when dr < 1 and dr is increasing, we can start applying fees at 0.95 rather than wait till we get to 1.


brandoniles (approved):
LGTM

